PR #2, opened 2022-07-29
Use Record<GenericObject> as output type for the Function

cbornet:


cbornet:
It makes the transformation function compatible only with LS2.10+ and Pulsar 2.11+.

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/Utils.java
```diff
@@ -57,6 +59,15 @@ public static TestTypedMessageBuilder<?> process(Record<GenericObject> record, T
     return context.getOutputMessage();
   }
 
+  public static Record<Object> process2(Record<GenericObject> record, TransformStep step)
```
aymkhalil, line 62: nit: Just use "process"

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformContext.java
```diff
@@ -76,7 +76,7 @@ public TransformContext(Context context, Object value) {
     }
   }
 
-  public void send() throws IOException {
+  public Record<Object> send() throws IOException {
```
aymkhalil, line 79: Any chance you could use a generic type here? `Record<T>`

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/Utils.java
cbornet: Oh yes, forgot to cleanup. Thanks for noticing.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformContext.java
cbornet: It must be the same type as TransformFunction's return type so I don't think we can make it a generic

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
```diff
@@ -79,7 +79,7 @@
  *     implementing the most common basic transformations</a>
  */
 @Slf4j
-public class TransformFunction implements Function<GenericObject, Void>, TransformStep {
+public class TransformFunction implements Function<GenericObject, Record<Object>>, TransformStep {
```
cbornet, line 82: Is it OK to return Record<Object> or should it be Record<GenericObject> ?

cbornet: @eolivelli

cbornet: Now returns `Record<GenericObject>`. Otherwise the PulsarSink starts a producer with Schema JSON.
We slip through Java generic type erasure to pass other types when we have to.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformContext.java
```diff
@@ -60,13 +60,12 @@ public TransformContext(Context context, Object value) {
       this.keySchema = kvSchema.getKeySchema();
       this.keyObject =
           this.keySchema.getSchemaInfo().getType() == SchemaType.AVRO
-              ? ((org.apache.pulsar.client.api.schema.GenericRecord) kv.getKey()).getNativeObject()
+              ? ((GenericObject) kv.getKey()).getNativeObject()
```
eolivelli, line 63: a "nativeObject" cannot be a "GenericObject" it is always:
- a primitive type
- a AVRO GenericRecord
- a JSON node
- a Protobuf something ???
- a KeyValue object

but NEVER a GenericObject or a Pulsar GenericRecord

this this will never happen in production, I am pretty sure

eolivelli (changes requested):

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformContext.java
cbornet: The GenericObject cast is on `kv.getKey()` not on the nativeObject

eolivelli: oh ! thanks ! you are right

eolivelli (approved):
LGTM

nice work!

PR #4, opened 2022-08-20
Implement flatten transform

aymkhalil:
First iteration of flatten transform. It works on the "key", "value", or both parts of the message. The delimiter used in this path is "_". Please note that in Avro, "." is not a valid field name: https://avro.apache.org/docs/1.11.1/specification/_print/#names

And example input is:
```JSON
{
    "level11": "key1",
    "level12": {
        "level21": "key2",
        "level22": {
            "level31": "key3",
            "level32": {
                "level41": "key4",
                "level42": "key5"
            }
        }
    }
}
```

After flattening becomes:
```JSON
{
  "level11": "key1",
  "level12_level21": "key2",
  "level12_level22_level31": "key3",
  "level12_level22_level32_level41": "key4" ,
  "level12_level22_level32_level41": "key5" 
}
```

This patch is working probably, but there are few items to follow up on (especially the first 1):
**To do list:** 
*  [Functionality: Blocker for Union types]Remove dependency on Avro's Field default value. It will not work with Union schemas "org.apache.avro.AvroTypeException: Invalid default for field a: "cloth" not a ["null","string"]"
  * [Performance] Cache schema to flattened schema
  * ~[DONE][UX] Make flatten delimiter configurable~
  * [UX] Validate flatten delimiter if possible
  * [Performance] Microbenchmark the flatten algorithm to optimize if needed
  * ~[DONE][Test Coverage] Add unit tests for schemas other than string for the unnested schema~
  * ~[DONE][Test Coverage] Add unit test for non-KeyValue schemas~
  * ~[DONE][Test Coverage] Add unit tests for Union schemas~
  * [e2e test] Add integration test

### DIFF
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/FlattenStep.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/FlattenStep.java
@@ -18,7 +18,9 @@ package com.datastax.oss.pulsar.functions.transforms;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import lombok.RequiredArgsConstructor;
+import lombok.Value;
 import org.apache.avro.generic.GenericData;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.pulsar.client.api.Schema;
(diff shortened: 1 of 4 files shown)


aymkhalil:
I tested this patch with 

```
bin/pulsar-admin functions localrun \
--jar ./pulsar-transformations/target/pulsar-transformations-1.0.3-SNAPSHOT.jar \
--classname com.datastax.oss.pulsar.functions.transforms.TransformFunction \
--tenant public \
--namespace default \
--name flatten116 \
--inputs public/default/flatten-input \
--output public/default/flatten-output \
--user-config "{\"steps\": [{\"type\": \"flatten\", \"delimiter\": \"_D_\", \"part\": \"key\"}]}"

```

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/FlattenStep.java
```diff
@@ -0,0 +1,165 @@
...
+      transformContext.setValueModified(true);
+    } else if (part.isPresent() && "key".equals(part.get())) {
+      validateAvro(transformContext.getKeySchema());
+      GenericRecord avroKeyRecord = (GenericRecord) transformContext.getKeyObject();
+      transformContext.setKeyObject(flattenGenericRecord(avroKeyRecord));
+      transformContext.setKeyModified(true);
+    } else if (part.isPresent() && "value".equals(part.get())) {
+      validateAvro(transformContext.getValueSchema());
+      GenericRecord avroValueRecord =
+          (transformContext.getValueObject()
```
aymkhalil, line 60: @cbornet when running from tests - if I pass a KV schema, I get the value as org.apache.avro.generic.GenericRecord, but if I pass a value only schema (key is a string), I get org.apache.pulsar.client.impl.schema.generic.GenericAvroRecord and I have to consult with the NativeObject.

I would expect both scenarios to give the same result (and that same result to be a GenericAvroRecrod). I'm not sure if this is a side effect [how the test triggers process ](https://github.com/datastax/pulsar-transformations/blob/c217ea692fcd2d67ed0b09bc34a8d0c8a9666def/pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/Utils.java#L68). 

WDYT?

eolivelli (approved):
LGTM

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/FlattenStep.java
```diff
@@ -0,0 +1,165 @@
...
+@RequiredArgsConstructor
+public class FlattenStep implements TransformStep {
+  // TODO: Cache schema to flattened schema
+  // TODO: Microbenchmark the flatten algorithm for performance optimization
+  // TODO: Validate flatten delimiter
+  // TODO: Add integration test
+
+  private static final String DEFAULT_DELIMITER = "_"; // '.' in not valid in AVRO field names
+
+  private final Optional<String> delimiter;
```
cbornet, line 38: Some people would say that `Optional` was only meant to be a type for return of methods and not for class attributes.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/FlattenStep.java
```diff
@@ -0,0 +1,165 @@
...
+    if (!part.isPresent()) {
+      validateAvro(transformContext.getKeySchema());
+      validateAvro(transformContext.getValueSchema());
+      GenericRecord avroKeyRecord = (GenericRecord) transformContext.getKeyObject();
+      GenericRecord avroValueRecord = (GenericRecord) transformContext.getValueObject();
+      transformContext.setKeyObject(flattenGenericRecord(avroKeyRecord));
+      transformContext.setValueObject(flattenGenericRecord(avroValueRecord));
+      transformContext.setKeyModified(true);
+      transformContext.setValueModified(true);
+    } else if (part.isPresent() && "key".equals(part.get())) {
```
cbornet, line 52: `part.isPresent()` is probably always true here ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/FlattenStep.java
cbornet: That's because of how the function framework works with GenericObject.
If you have a simple Schema.AVRO then the pulsar.GenericRecord you get is a GenericAvroRecord and the wrapped nativeObject is the avro.GenericRecord.
If you have a KV, then the pulsar.GenericRecord is a GenericObjectWrapper and the wrapped nativeObject is a KV<pulsar.GenericRecord, pulsar.GenericRecord>. So when initializing the TransformContext, we unwrap again the key and value.
So I think this code is not useful for real usage, the valueObject will always be avro.GenericRecord. Or did you have issues while testing in integration/live ?

cbornet:
Nice work ! I left a couple of comments.
Integration tests could validate the overall working of the step. Especially validate this https://github.com/datastax/pulsar-transformations/pull/4#discussion_r954130988

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/FlattenStep.java
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  void validateAvro(Schema<?> schema) {
+    if (schema == null) {
+      throw new IllegalStateException("Flatten requires non-null schemas!");
+    }
+
+    if (schema.getSchemaInfo().getType() != SchemaType.AVRO) {
+      throw new IllegalStateException(
+          "Unsupported schema type for Flatten: " + schema.getSchemaInfo().getType());
```
cbornet, line 81: This could be covered by a unit test ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/FlattenStep.java
aymkhalil: you are right. let me change this in-light of the other comment about using optional as class member.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/FlattenStep.java
aymkhalil: makes sense, let me find a better way to handle it.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/FlattenStep.java
aymkhalil: no you are right. In integ, I'll always get avro.GenericRecord, but with unit testing, I get the described behavior above. Ideally the code path should be the same for unit/integ testing, let me see if it is possible (and then I can remove the if/else part)

PR #6, opened 2022-08-31
Add JMS filters based predicate to transformation steps

aymkhalil:
This is the first iteration of adding JSM filters as predicates to enable/disable a specific step. The idea is to add an optional filter config to each step, for example:

```
"{\"steps\": 
   [{\"type\": \"drop-fields\", \"fields\": \"body\", 
     \"filter\":\"value.title = 'title' AND key.id = 'title' AND property LIKE '%prop%'\"}]}"
```

Fields in the predicate can be prefixed by key or value to access the respective field from the generic record. If no prefix is provided, the field name applies by default to the custom message properties provided.

TODO:
* The JMS selector doesn't accept a "."  in the field name. A workaround is implement to replace "." with "$" just to showcase the complete use case. I think we should try to avoid parsing the SQL predicate ourselves to preprocess the field names. Alternative options is to have the user pass the scope of the fields separately on the step config. 
* Integ tests is not yet possible, when running the functions in localrun mode, it fails to load the JMS dependecies: 
```
java.lang.NoClassDefFoundError: javax/jms/JMSException
```
* Message header matching other than the user defined props

### DIFF
--- a/pulsar-transformations/pom.xml
+++ b/pulsar-transformations/pom.xml
@@ -72,6 +72,21 @@
       <scope>provided</scope>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>${pulsar.groupId}</groupId>
+      <artifactId>pulsar-jms-filters</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <!--    <dependency>-->
+    <!--      <groupId>javax.resource</groupId>-->
+    <!--      <artifactId>javax.resource-api</artifactId>-->
+    <!--      <scope>provided</scope>-->
+    <!--    </dependency>-->
+    <!--    <dependency>-->
+    <!--      <groupId>jakarta.jms</groupId>-->
+    <!--      <artifactId>jakarta.jms-api</artifactId>-->
+    <!--      <scope>provided</scope>-->
+    <!--    </dependency>-->
   </dependencies>
   <build>
     <plugins>
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/JmsPredicate.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/JmsPredicate.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.pulsar.functions.transforms;
+
+import com.datastax.oss.pulsar.jms.selectors.SelectorSupport;
+import java.util.function.Predicate;
+import javax.jms.JMSException;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.pulsar.client.api.Message;
+
+public class JmsPredicate implements Predicate<TransformContext> {
+
+  private final SelectorSupport selector;
+
+  public JmsPredicate(String filter) {
+    try {
+      // A hacky way in to init the predicate as JMS filters doesn't accept . in field names,
+      // Parsing and passing a positional list of ket/value is tricky as we need to understand the
+      // SQL syntax
+      // TODO: Find a better way to handle the scope of the field
+      String sanitizedFilter =
+          filter.replaceAll("key\\.", "key\\$").replaceAll("value\\.", "value\\$");
+      selector = SelectorSupport.build(sanitizedFilter, true);
+    } catch (JMSException ex) {
+      throw new IllegalArgumentException("invalid filter: " + filter, ex);
+    }
+  }
+
+  @Override
+  public boolean test(TransformContext transformContext) {
+    try {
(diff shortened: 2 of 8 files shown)


pulsar-transformations/pom.xml
```diff
@@ -72,6 +72,21 @@
       <scope>provided</scope>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>${pulsar.groupId}</groupId>
```
cbornet, line 76:
```suggestion
      <groupId>com.datastax.oss</groupId>
```
This is not from Pulsar core. It's a datastax only project so we won't switch between forks.

pulsar-transformations/pom.xml
```diff
@@ -72,6 +72,21 @@
...
+      <artifactId>pulsar-jms-filters</artifactId>
+      <scope>provided</scope>
```
cbornet, line 78: Need to provide the version in dependencyManagement

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/JmsPredicate.java
```diff
@@ -0,0 +1,72 @@
...
+    String scope = null;
+    String key = field;
+    if (field.startsWith("key$") || field.startsWith("value$")) {
+      String[] filterParts = field.split("\\$", 2);
+      scope = filterParts[0];
+      key = filterParts[1];
+    }
+
+    if (scope == null) {
+      Message message = context.getContext().getCurrentRecord().getMessage().orElse(null);
```
cbornet, line 61: If we do in the future Source+Function, there's a high probability that the message is not present.
Will it be an issue ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/JmsPredicate.java
```diff
@@ -0,0 +1,72 @@
...
+      return message == null ? null : message.getProperties().get(key);
```
cbornet, line 62: Would replace by
```java
return context.getContext().getCurrentRecord().getMessage()
          .map(m -> m.getProperties().get(key))
          .orElse(null);
```
Just need to make `key` effectively final but that's not hard.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
```diff
@@ -151,7 +161,12 @@ public Void process(GenericObject input, Context context) throws Exception {
...
   @Override
   public void process(TransformContext transformContext) throws Exception {
-    for (TransformStep step : steps) {
+    for (StepPredicatePair pair : steps) {
+      TransformStep step = pair.getTransformStep();
+      Predicate<TransformContext> predicate = pair.getPredicate();
+      if (predicate != null && !pair.getPredicate().test(transformContext)) {
+        continue;
+      }
       step.process(transformContext);
```
cbornet, line 170:
```suggestion
      if (predicate == null || predicate.test(transformContext)) {
        step.process(transformContext);
      }
```

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/FiltersTest.java
```diff
@@ -0,0 +1,43 @@
...
+                case "prop1":
+                  return "my-prop1";
+                case "prop2":
+                  return "my-prop12";
+                default:
+                  return "";
+              }
+            });
+
+    System.out.println(m);
```
cbornet, line 41: Use a logger ?

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/JmsPredicateTest.java
```diff
@@ -0,0 +1,106 @@
...
+import org.testng.annotations.Test;
+
+public class JmsPredicateTest {
+
+  @Test
+  void testKeyScope() {
+    JmsPredicate match = new JmsPredicate("key.keyField1='key1' AND key.keyField2='key2'");
+    JmsPredicate matchLike = new JmsPredicate("key.keyField1 LIKE '%k%'");
+    JmsPredicate dontMatch = new JmsPredicate("value.keyField1 ='key1'");
+    JmsPredicate dontMatchLike = new JmsPredicate("key.keyField1 LIKE '%value%'");
```
cbornet, line 33: Use parameterized tests ?

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/TransformFunctionTest.java
```diff
@@ -127,6 +142,90 @@ void testDropFields() throws Exception {
...
+    Map<String, Object> config =
+        new Gson().fromJson(userConfig, new TypeToken<Map<String, Object>>() {}.getType());
+    TransformFunction transformFunction = new TransformFunction();
+
+    Record<GenericObject> record = Utils.createTestAvroKeyValueRecord();
+    Utils.TestContext context = new Utils.TestContext(record, config);
+    transformFunction.initialize(context);
+    transformFunction.process(record.getValue(), context);
+
+    Utils.TestTypedMessageBuilder<?> message = context.getOutputMessage();
```
cbornet, line 162: This doesn't work anymore. Need to work on the Record returned by `transformFunction.process`

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/JmsPredicate.java
cbornet: I think you should get the properties from the context instead

aymkhalil:
Closing in favor of https://github.com/datastax/pulsar-transformations/pull/9

PR #9, opened 2022-09-04
Implement jstl based step predicates 

aymkhalil:
This is a Unified Expression Language (EL) based step predicates implementation. Please check #6 for comparison with JMS based 
filter implementation. A sample when condition would be:

`${key.field1 == 'value' && key.recordField.field2.toUpperCase == 'VALUE2"`

Few implementation details:
* Each step has an option `when` config. It is written in the EL defined [here](https://javaee.github.io/tutorial/jsf-el001.html#BNAHQ)
* The chosen java implementation of the EL is [JUEL](https://github.com/beckchr/juel)
* The language natively supports nested props access via the dot notation. 
* In order to generate bindings for the fields in the expression, and given the dynamic nature of our generic records and k/v schema, a lazy map is initialized with an adapter object to expose parts of the TransformContext as request by the expression evaluater. Note that an alternative to this would've been to generate java beans props dynamically at runtime (seems possible via spring's [BeanDefinitionRegistryPostProcessor](https://docs.spring.io/spring-framework/docs/3.0.x/javadoc-api/org/springframework/beans/factory/support/BeanDefinitionRegistryPostProcessor.html) 
* The user can choose between 2 different scopes : `[key, value]`. Header fields are avaible as top level attributes like`topicName`. Please note that the header `key` (aka. the partition key) is called `messageKey` to avoid conflict with an INLINE KV key if one exists
* The user defined props can be accessed via `${header.properties.propertyName}` syntax
* Header props are enabled on an `allowlisting` bases so we have control over which headers are included the headers. 
* Adding new predicate types (if needed, say we want to give our users flexibility to choose between jstl and jms Language), an implementation of the `TransformPredicate` is all what's needed. The a factory

If this is the right direction (vs the JMS one) - I'll use this PR as basis for the final implementation (mainly adding more test coverage)

### DIFF
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
@@ -15,6 +15,8 @@
  */
 package com.datastax.oss.pulsar.functions.transforms;
 
+import com.datastax.oss.pulsar.functions.transforms.predicate.StepPredicatePair;
+import com.datastax.oss.pulsar.functions.transforms.predicate.jstl.JstlPredicate;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
 import java.util.ArrayList;
@@ -23,6 +25,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Predicate;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.pulsar.client.api.schema.GenericObject;
 import org.apache.pulsar.common.schema.SchemaType;
@@ -77,7 +80,7 @@ import org.apache.pulsar.functions.api.Record;
  *       "type": "cast", "schema-type": "STRING"
  *     },
  *     {
- *       "type": "flatten", "delimiter" : "_" "part" : "value"
+ *       "type": "flatten", "delimiter" : "_" "part" : "value", "when": "$(value.field == 'value')"
  *     }
  *   ]
  * }
@@ -90,7 +93,7 @@ import org.apache.pulsar.functions.api.Record;
 public class TransformFunction
     implements Function<GenericObject, Record<GenericObject>>, TransformStep {
 
-  private final List<TransformStep> steps = new ArrayList<>();
+  private final List<StepPredicatePair> steps = new ArrayList<>();
   private final Gson gson = new Gson();
 
   @Override
@@ -109,25 +112,29 @@ public class TransformFunction
     }
     for (Map<String, Object> step : stepsConfig) {
       String type = getRequiredStringConfig(step, "type");
+      Optional<String> when = getStringConfig(step, "when");
+      Predicate<TransformContext> predicate = when.map(JstlPredicate::new).orElse(null);
+      TransformStep transformStep;
       switch (type) {
         case "drop-fields":
-          steps.add(newRemoveFieldFunction(step));
+          transformStep = newRemoveFieldFunction(step);
           break;
         case "cast":
-          steps.add(newCastFunction(step));
+          transformStep = newCastFunction(step);
           break;
         case "merge-key-value":
-          steps.add(new MergeKeyValueStep());
+          transformStep = new MergeKeyValueStep();
           break;
         case "unwrap-key-value":
-          steps.add(newUnwrapKeyValueFunction(step));
+          transformStep = newUnwrapKeyValueFunction(step);
           break;
         case "flatten":
-          steps.add(newFlattenFunction(step));
+          transformStep = newFlattenFunction(step);
           break;
         default:
           throw new IllegalArgumentException("invalid step type: " + type);
       }
+      steps.add(new StepPredicatePair(transformStep, predicate));
     }
   }
 
@@ -151,7 +158,12 @@ public class TransformFunction
 
   @Override
   public void process(TransformContext transformContext) throws Exception {
-    for (TransformStep step : steps) {
+    for (StepPredicatePair pair : steps) {
+      TransformStep step = pair.getTransformStep();
+      Predicate<TransformContext> predicate = pair.getPredicate();
+      if (predicate != null && !predicate.test(transformContext)) {
+        continue;
+      }
       step.process(transformContext);
     }
   }
@@ -196,7 +208,14 @@ public class TransformFunction
 
   public static FlattenStep newFlattenFunction(Map<String, Object> step) {
     FlattenStep.FlattenStepBuilder builder = FlattenStep.builder();
-    getStringConfig(step, "part").ifPresent(builder::part);
+    getStringConfig(step, "part")
+        .ifPresent(
+            part -> {
+              if (!("key".equals(part) || "value".equals(part))) {
+                throw new IllegalArgumentException("invalid 'part' parameter: " + part);
+              }
+              builder.part(part);
+            });
     getStringConfig(step, "delimiter").ifPresent(builder::delimiter);
     return builder.build();
   }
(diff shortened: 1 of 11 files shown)


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/predicate/StepPredicatePair.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.datastax.oss.pulsar.functions.transforms.TransformContext;
+import com.datastax.oss.pulsar.functions.transforms.TransformStep;
+import java.util.function.Predicate;
+import lombok.Data;
+
+@Data
+public class StepPredicatePair {
+  private final TransformStep transformStep;
+  private final Predicate<TransformContext> predicate;
```
eolivelli, line 26: This is very interesting.

It is not obvious if the predicate is evaluated on the initial message or on the intermediate record.

I think that current implementation, based on the current status of the Record is good but we have to document it clearly

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/predicate/jstl/JstlTransformContextAdapter.java
```diff
@@ -0,0 +1,110 @@
...
+      };
+
+  private final Map<String, Object> lazyValue = LazyMap.lazyMap(new HashMap<>(), valueTransformer);
+
+  /** A header transformer to return message headers the user is allowed to filter on. */
+  private Transformer<String, Object> headerTransformer =
+      (fieldName) -> {
+        Record<?> currentRecord = transformContext.getContext().getCurrentRecord();
+        return currentRecord
+            .getMessage()
```
eolivelli, line 72: The message is not always available, it is available only in the first step.
We should read topic name and properties and all the other things from the TransformContext

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/TransformFunctionTest.java
```diff
@@ -126,6 +139,90 @@ void testDropFields() throws Exception {
     assertNull(valueAvroRecord.getSchema().getField("valueField3"));
   }
 
+  @Test
+  void testMatchingPredicate() throws Exception {
+    String userConfig =
+        (""
+            + "{\"steps\": ["
+            + "    {\"type\": \"drop-fields\", \"fields\": \"keyField1\", \"when\": \"${key.keyField1 == 'key1'}\"},"
```
eolivelli, line 147: I think that the user should not write ${ and }. We can add them automatically

eolivelli (commented):
I think that we are on our way.
I left some comments, especially the fact that the user should not use the ${} syntax but only write the expression

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
```diff
@@ -151,7 +158,12 @@ public Record<GenericObject> process(GenericObject input, Context context) throw
 
   @Override
   public void process(TransformContext transformContext) throws Exception {
-    for (TransformStep step : steps) {
+    for (StepPredicatePair pair : steps) {
+      TransformStep step = pair.getTransformStep();
+      Predicate<TransformContext> predicate = pair.getPredicate();
+      if (predicate != null && !predicate.test(transformContext)) {
+        continue;
```
cbornet, line 165: Should reverse this condition and avoid the `continue`

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/predicate/StepPredicatePair.java
cbornet: Agreed that the predicate should apply to the current status.

aymkhalil: Makes sense. I'm not sure if it clarifies, but I can move the `predicate` to the `TransformStep` class so that it is more encapsulated in the step and since the step operates on the current state, so will the predicate.  

The only reason I kept them separate (and grouped in pairs) is because the step actually has nothing to do with the predicate, it is the higher level processor that takes a "process, don't process" decision. Added predicate evaluation function on something like an "AbstractTransformStep" sounded like an overkill... Will add the java docs for now.

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/TransformFunctionTest.java
aymkhalil: Sure. on that note, should we mention in out public docs that the we use expression language behind the scenes? 

We need to either do that (so user know which syntax to use inside the ${}) OR document supported syntax ourselves (which is a subset of UEL)

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
aymkhalil: sure, I wanted to use the filter syntax in streams to make it cleaner, but it forces us to handle the checked exception in-place...

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/predicate/jstl/JstlTransformContextAdapter.java
```diff
@@ -0,0 +1,133 @@
...
+   * value or another lazy map for nested evaluation.
+   */
+  private Transformer<String, Object> keyTransformer =
+      (fieldName) -> {
+        if (this.transformContext.getKeyObject() instanceof GenericRecord) {
+          GenericRecord genericRecord = (GenericRecord) this.transformContext.getKeyObject();
+          JstlPredicate.GenericRecordTransformer transformer =
+              new JstlPredicate.GenericRecordTransformer(genericRecord);
+          return transformer.transform(fieldName);
+        } else if (this.transformContext.getKeyObject() instanceof KeyValue) {
```
eolivelli, line 45: this is not needed, we can drop this case, it is not possible in Pulsar

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/predicate/jstl/JstlTransformContextAdapter.java
```diff
@@ -0,0 +1,133 @@
...
+   * a value or another lazy map for nested evaluation.
+   */
+  private Transformer<String, Object> valueTransformer =
+      (fieldName) -> {
+        if (this.transformContext.getValueObject() instanceof GenericRecord) {
+          GenericRecord genericRecord = (GenericRecord) this.transformContext.getValueObject();
+          JstlPredicate.GenericRecordTransformer transformer =
+              new JstlPredicate.GenericRecordTransformer(genericRecord);
+          return transformer.transform(fieldName);
+        } else if (this.transformContext.getValueObject() instanceof KeyValue) {
```
eolivelli, line 66: this is not needed, it is not possible in Pulsar

eolivelli (approved):
LGTM

I would drop the useless code, apart from that it looks great !

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/predicate/jstl/JstlTransformContextAdapter.java
aymkhalil: removed as per the offline discussion

cbornet:
LGTM. Awesome work !

cbornet (approved):

PR #12, opened 2022-09-15
Add config validation with Json Schema

cbornet:


aymkhalil:
No an expert here but reading this https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/, it doesn't seem discriminator are meant for validation but rather a hint to consumers on which property to use as a type to distinguish objects in case of `anyOf` or `anyOf`.

Is it possible to avoid using OpenAPI stuff and stick to vanilla json schema validations for simplicity?

pulsar-transformations/src/main/resources/schema.json
```diff
@@ -0,0 +1,180 @@
+{
```
aymkhalil, line 1: maybe name the file `config-schema.json`

cbornet:
I use the OAS discriminator to get the relevant error message for the user.
Otherwise as all models from the oneOf/anyOf don't match, the order of the validation errors is not defined.
By using a discriminator, the first validation error is the one from the model that matched the discriminator and that's the one I report to the user (see `errors.stream().findFirst()`).
That said, the discriminator doesn't work correctly with the lib I use so I had to implement my own model detection and it could be removed. I think we can keep it for users that use libs that implement it correctly. There's no harm as other libs will just ignore it.

pulsar-transformations/src/main/resources/schema.json
cbornet: sure

aymkhalil (approved):

PR #13, opened 2022-09-15
Change the fields field in drop-fields step by a list of strings

cbornet:


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
```diff
@@ -168,8 +168,18 @@ public void process(TransformContext transformContext) throws Exception {
   }
 
   public static DropFieldStep newRemoveFieldFunction(Map<String, Object> step) {
-    String fields = getRequiredStringConfig(step, "fields");
-    List<String> fieldList = Arrays.asList(fields.split(","));
+    List<String> fieldList = new ArrayList<>();
+    Object fields = step.get("fields");
+    if (fields instanceof List) {
+      for (Object field : (List<?>) fields) {
+        if (field instanceof String && !((String) field).isEmpty()) {
```
aymkhalil, line 175: I think this is ok here, with json schema validation, non-string fields names should not pass the validation right?

aymkhalil (approved):
LGTM assuming json validation with catch non-string fields.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
cbornet: Yes, if we use Json Schema validation, we can remove this block and directly cast to `List<String>`

PR #14, opened 2022-09-15
Implement drop transform step

aymkhalil:
This step will drop the current record from further processing. It is most useful when used with `when` predicates to selectively choose which messages to drop.

Step config example:

```
{
    "steps": [
        {
            "type": "drop-message",
            "when": "messageKey=='abc'"
        }
    ]
}
``` 

**Implementation note:**
There is a slight difference between the handling of null [Function.process](https://github.com/apache/pulsar/blame/master/pulsar-functions/api-java/src/main/java/org/apache/pulsar/functions/api/Function.java#L40) result between LS and OSS Pulsar - although both lead to an auto ack of the message without further processing [[LS](https://github.com/datastax/pulsar/blob/b980f73909c00997b5c5c97623f470aa65de1cdc/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstanceRunnable.java#L396), [OSS](https://github.com/apache/pulsar/blob/master/pulsar-functions/instance/src/main/java/org/apache/pulsar/functions/instance/JavaInstanceRunnable.java#L392)]. The drop step in this PR relies on the fact that `null` means ignore the message from further processing.

### DIFF
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformContext.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformContext.java
@@ -48,6 +48,7 @@ public class TransformContext {
   private String key;
   private Map<String, String> properties;
   private String outputTopic;
+  private boolean dropCurrentRecord;
 
   public TransformContext(Context context, Object value) {
     Record<?> currentRecord = context.getCurrentRecord();
@@ -76,6 +77,9 @@ public class TransformContext {
   }
 
   public Record<GenericObject> send() throws IOException {
+    if (dropCurrentRecord) {
+      return null;
+    }
     if (keyModified
         && keySchema != null
         && keySchema.getSchemaInfo().getType() == SchemaType.AVRO) {
--- a/pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/TransformFunctionTest.java
+++ b/pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/TransformFunctionTest.java
@@ -36,6 +36,7 @@ import org.apache.pulsar.common.schema.SchemaInfo;
 import org.apache.pulsar.common.schema.SchemaType;
 import org.apache.pulsar.functions.api.Context;
 import org.apache.pulsar.functions.api.Record;
+import org.testng.Assert;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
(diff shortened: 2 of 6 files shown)


pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/TransformFunctionTest.java
```diff
@@ -223,6 +230,59 @@ void testMixedPredicate() throws Exception {
...
+    SchemaInfo schemaInfo = recordSchemaBuilder.build(SchemaType.AVRO);
+    GenericSchema<GenericRecord> genericSchema = Schema.generic(schemaInfo);
+
+    GenericRecord genericRecord =
+        genericSchema.newRecordBuilder().set("firstName", "Galina").build();
+
+    Record<GenericObject> record = new Utils.TestRecord<>(genericSchema, genericRecord, "test-key");
+    String userConfig =
+        (""
+            + "{\"steps\": ["
```
eolivelli, line 272: It seems to me that we don't have a test about having the drop step in the middle of the pipeline
What about adding one?

eolivelli (approved):

tests/src/test/java/com/datastax/oss/pulsar/functions/transforms/tests/AbstractDockerTest.java
```diff
@@ -243,12 +244,41 @@ public void testKVAvroWhenPredicate() throws Exception {
     assertEquals(keyValue.getValue().getNativeObject().toString(), "{\"d\": \"d\"}");
   }
 
+  @Test
+  void testDropMessageOnPredicateMatch() throws Exception {
+    String userConfig =
+        ("" + "{\"steps\": [" + "    {\"type\": \"drop\", \"when\": \"value.a=='a'\"}" + "]}");
```
cbornet, line 250: Make only only one string ?

tests/src/test/java/com/datastax/oss/pulsar/functions/transforms/tests/AbstractDockerTest.java
```diff
@@ -243,12 +244,41 @@ public void testKVAvroWhenPredicate() throws Exception {
...
+    GenericRecord value =
+        testTransformFunction(
+            userConfig, Schema.AVRO(Pojo1.class), new Pojo1("a", "b"), null, true);
+    assertNull(value);
+  }
+
+  @Test
+  void testDropMessageOnPredicateMissMatch() throws Exception {
+    String userConfig =
+        ("" + "{\"steps\": [" + "    {\"type\": \"drop\", \"when\": \"value.a=='a'\"}" + "]}");
```
cbornet, line 260: Make only only one string ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformContext.java
```diff
@@ -48,6 +48,7 @@ public class TransformContext {
   private String key;
   private Map<String, String> properties;
   private String outputTopic;
+  private boolean dropCurrentRecord;
```
cbornet, line 51: What if we set the value to null instead of adding a field ?

cbornet (approved):

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/TransformFunctionTest.java
aymkhalil: +1

tests/src/test/java/com/datastax/oss/pulsar/functions/transforms/tests/AbstractDockerTest.java
aymkhalil: sure

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformContext.java
aymkhalil: In KV schemas for example, Value is allowed to be null. We use this in CDC to send a mutation to indicate that a record was deleted (so value would be null, but still a record would be sent). 

I thought about setting both key and value to null to indicate the delete, but found the flag more explicit.

cbornet: We should have some tests with KV with null key or value.

PR #17, opened 2022-09-19
Add docs for conditional steps (when step config)

aymkhalil:
Add docs for `when` step configs (and the `drop` transform)  #15


cbornet:
LGTM.
Can you resolve the conflict ?

PR #19, opened 2022-09-19
Implement compute-fields step

aymkhalil:
This patch adds support for `compute-fields` transformation step.

Each `compute-fields` has an a list of `fields` to compute. Each `field` is a map of the following mandatory fields:
* `name`[mandatory]: The name of the new field. If the field already exist, it will be overwritten.
* `expression`[mandatory]: a JSTL based expression that computes the value of the fields. It could evaluate the fields based on operators only (like `1 + 1`) or it could reference fields from other parts of the message via `key.` or `value.` prefixes 
* `type`[mandatory]: The expected type of the evaluated expression and the thew new fields.  The following types are the only supported ones in this patch `[STRING, INT32, INT64, FLOAT, DOUBLE, BOOLEAN]`
* `part` [optional]: This config decides where to add fields (key or value). If not set, it computes the field on value.


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
```diff
@@ -0,0 +1,153 @@
...
+      newRecord.put(field.getName(), field.getEvaluator().evaluate(context));
+    }
+    return newRecord;
+  }
+
+  private Schema getAvroSchema(ComputeFieldType type) {
+    switch (type) {
+      case STRING:
+        return fieldTypeToAvroSchemaCache.computeIfAbsent(
+            type, (ignored) -> Schema.create(Schema.Type.STRING));
```
eolivelli, line 133: do we need to create new instances ? or can we cache the basic schemas ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
```diff
@@ -290,6 +295,22 @@ public static FlattenStep newFlattenFunction(JsonNode node) {
...
+    List<ComputeField> fieldList = new ArrayList<>();
+    node.get("fields")
+        .iterator()
+        .forEachRemaining(
+            it ->
+                fieldList.add(
+                    ComputeField.builder()
+                        .name(it.get("name").asText())
+                        .expression(it.get("expression").asText())
+                        .type(ComputeFieldType.valueOf(it.get("type").asText()))
```
eolivelli, line 308: we must validate  that the "type" is a supported type, this will prevent runtime errors.
also, the error should report the list of available times

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlTransformContextAdapter.java
```diff
@@ -115,4 +113,28 @@ public Object getValue() {
...
+    public GenericRecordTransformer(GenericRecord genericRecord) {
+      this.genericRecord = genericRecord;
+    }
+
+    @Override
+    public Object transform(String key) {
+      Object value = null;
+      if (genericRecord.hasField(key)) {
+        value = genericRecord.get(key);
+        if (value instanceof GenericRecord) {
```
eolivelli, line 131: for the future:
what about "arrays" ?
this is not top priority, but we should at least create a GH issue

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/predicate/JstlPredicate.java
```diff
@@ -0,0 +1,46 @@
...
+      throw new IllegalArgumentException("invalid when: " + when, ex);
+    }
+  }
+
+  @Override
+  public boolean test(TransformContext transformContext) {
+    try {
+      return this.evaluator.evaluate(transformContext);
+    } catch (PropertyNotFoundException ex) {
+      log.warn("a property in the when expression was not found in the message", ex);
```
eolivelli, line 42: can we provide more details, like the name of the property ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/ComputeField.java
```diff
@@ -0,0 +1,76 @@
...
+        case INT64:
+          return long.class;
+        case FLOAT:
+          return float.class;
+        case DOUBLE:
+          return double.class;
+        case BOOLEAN:
+          return boolean.class;
+        default:
+          throw new UnsupportedOperationException("Unsupported compute field type: " + type);
```
eolivelli, line 72: can we report the list of supported types ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/ComputeFieldType.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package com.datastax.oss.pulsar.functions.transforms.model;
+
+public enum ComputeFieldType {
+  STRING,
+  INT32,
+  INT64,
+  FLOAT,
+  DOUBLE,
+  BOOLEAN
```
eolivelli, line 24: can we create an issue to support date/times ?

eolivelli (changes requested):
I have left some comments.

I have a higher level question:
for AVRO, we are adding a new field, with a given Schema,
will the field be "NULLABLE" ?
I think that we should take into consideration nulls, both as return value.

I think that the safest behaviour would be to always create "NULLABLE" fields
or we can add a configuration entry:
- nullable: true|false (default: true)

if nullable:
- the schema is UNION(NULL, TYPE)
- nulls are alllowed

if non-nullable
- the schema is TYPE
- nulls throw a runtimeerror (the message will be retried)

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
```diff
@@ -0,0 +1,153 @@
...
+    }
+    // Add computed fields
+    for (ComputeField field : fields) {
+      newRecord.put(field.getName(), field.getEvaluator().evaluate(context));
+    }
+    return newRecord;
+  }
+
+  private Schema getAvroSchema(ComputeFieldType type) {
+    switch (type) {
```
cbornet, line 130: Seems that you can map the ComputeFieldType to Schema.Type and do the `fieldTypeToAvroSchemaCache.computeIfAbsent(
            type, (ignored) -> Schema.create(Schema.Type.STRING))` after the `switch`

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
cbornet: I think it's validated by the Json Schema. And the list of possible is in the error message. @aymkhalil do you confirm ?

pulsar-transformations/src/main/resources/config-schema.yaml
```diff
@@ -141,6 +142,57 @@ components:
...
+        - "$ref": "#/components/schemas/Step"
+        - type: object
+          description: Adds or updates fields dynamically based on an expression (Currently only AVRO is supported).
+          properties:
+            type:
+              type: string
+              enum:
+                - compute-fields
+            fields:
+              type: [ array, 'null' ]
```
cbornet, line 156: I think we should not allow `null` ?

pulsar-transformations/src/main/resources/config-schema.yaml
```diff
@@ -141,6 +142,57 @@ components:
...
+              items:
+                type:
+                  - object
+                minLength: 1
```
cbornet, line 160: minLength is for string types.
It doesn't make sense here

pulsar-transformations/src/main/resources/config-schema.yaml
```diff
@@ -141,6 +142,57 @@ components:
...
+                properties:
+                  name:
+                    type: string
+                    description: The name of the field to compute. If the field name already exist, it will be overwritten.
```
cbornet, line 164: we can add a check that the name is not empty

pulsar-transformations/src/main/resources/config-schema.yaml
```diff
@@ -141,6 +142,57 @@ components:
...
+                  expression:
+                    type: string
```
cbornet, line 166: do we need to check that the expression is not empty ?

pulsar-transformations/src/main/resources/config-schema.yaml
```diff
@@ -141,6 +142,57 @@ components:
...
+                    description: An EL-like expression to compute the field value. It can refer to other attributes in the record.
+                  type:
+                    type: string
```
cbornet, line 169: do we allow null ?

pulsar-transformations/src/main/resources/config-schema.yaml
```diff
@@ -141,6 +142,57 @@ components:
...
+                    description: The data type of the computed fields. It should match with the output of the expression.
+                    enum:
+                      - STRING
+                      - INT32
+                      - INT64
+                      - FLOAT
+                      - DOUBLE
+                      - BOOLEAN
+                  part:
+                    type: string
```
cbornet, line 179: Maybe reuse the `Part` definition ?

pulsar-transformations/src/main/resources/config-schema.yaml
```diff
@@ -141,6 +142,57 @@ components:
...
+                    description: The part of the message to apply the computed fields to. If left empty, the fields will apply to the `value`.
```
cbornet, line 180: For the other transforms, we apply on both key and value if part is empty. Shouldn't we do the same for consistency ?

pulsar-transformations/src/main/resources/config-schema.yaml
```diff
@@ -141,6 +142,57 @@ components:
...
+                    enum:
+                      - key
+                      - value
+                    default: value
+                description: The list of fields to compute
+                message:
+                  minLength: "field name in 'fields' must not be empty"
```
cbornet, line 187: This is to override the default error message. Here minLength doesn't apply to type object so this part shall be removed.

pulsar-transformations/src/main/resources/config-schema.yaml
```diff
@@ -141,6 +142,57 @@ components:
...
+                required:
+                  - name
+                  - expression
+                  - type
```
cbornet, line 191: Is type required or could it be inferred ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
cbornet: See https://github.com/datastax/pulsar-transformations/pull/19/files#diff-02d4ba68ec7c9d19097b2324e4a85fdf66c1f1d9266a976fa3e905d2f1f1ba01R129

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
aymkhalil: It is already cached. `fieldTypeToAvroSchemaCache` is a map and the schema is computed on the first time only via(`computeIfAbsent`)

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
aymkhalil: Confirmed. All validation logic we get out of the box from the JSON schema (and tested in unit tests). Thanks @cbornet for switching to schema based config, it takes away the validation burden!

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlTransformContextAdapter.java
aymkhalil: This code in particular is to support nested access. Arrays already supported by the syntax. I will few unit tests for it in the next patch.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/predicate/JstlPredicate.java
aymkhalil: It is already part of the EL ex. The exception msg will contain smth like:
```
javax.el.PropertyNotFoundException: Cannot find property xyz
```

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/ComputeField.java
aymkhalil: we can, but the full list of types is completely validated at step build time via the config schema validation. As far as customers are concerned, they'll get a meaningful error earlier with supported types enumerated, smth like:
```
Invalid 'compute-fields' step config: $.fields[0].type: does not have a value in the enumeration [STRING, INT32, INT64, FLOAT, DOUBLE, BOOLEAN]
```

This part of the code is just a safety net. I think the config validation is sufficient to avoid updating the types in two places every time we add support for a new one.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/ComputeFieldType.java
aymkhalil: https://github.com/datastax/pulsar-transformations/issues/20

pulsar-transformations/src/main/resources/config-schema.yaml
aymkhalil: ops, removing and adding test

pulsar-transformations/src/main/resources/config-schema.yaml
aymkhalil: Yes! I wanted to restrict array length, not item :(. Fixing with `minItems`

pulsar-transformations/src/main/resources/config-schema.yaml
aymkhalil: not really, this should be [key, value] or not provided. null will override the default value which is value

aymkhalil: hmm, I wonder what is the use case? Will leave it out for now.

pulsar-transformations/src/main/resources/config-schema.yaml
aymkhalil: Yeah check the comment above, from UX, when computing a field, generally I don't think the same field need to be created in both parts. In this case, I'm preferring usability over consistency.

pulsar-transformations/src/main/resources/config-schema.yaml
aymkhalil: In the current implementation it is explicit. I think inference will open the door for some issues and would prefer predictable from the user on the schema. For example if I user types `${1}` and we bind all variables to `Object.class`:
```
ValueExpression e = factory.createValueExpression(context, "${1}", Object.class); 
System.out.println(e.getValue(context).getClass().getName()); 

OUTPUT:
java.lang.Long
```

You'll get a long, but int would be a better fit. We an alway do our own inference like to check if this long can be stored in int and use int, but it can get tricky with other types. Also I'm thinking, if one day we switch the underlying JSTL engine, it is hard to guarantee it'll return the same types...

aymkhalil:
> I think that the safest behavior would be to always create "NULLABLE" fields or we can add a configuration entry:

Agreed. Let me add an `optional` or `nullable` boolean config and have the user decide ...

pulsar-transformations/src/main/resources/config-schema.yaml
aymkhalil: So I'll make the field optional (so a union of nullable, type) as per Enrico's. So the user still has to provide an explicit type, but if the fields evaluates to null, it will not break their schema.

cbornet:
I've been thinking: instead of specifying a `part`, maybe we could do `key.newField1` and `value.newField2` in the value ?
This could be very powerful as in the future we could also have `topicName` (routing !), `messageKey` or `property.newProp`

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
```diff
@@ -290,6 +295,24 @@ public static FlattenStep newFlattenFunction(JsonNode node) {
...
+                        .part(it.get("part") == null ? null : it.get("part").asText())
+                        .optional(
+                            it.get("optional") == null ? false : it.get("optional").asBoolean())
```
eolivelli, line 311: I would set it "optional" by default, it is easier, because if it is optional you don't care about the fact that you have nulls or not

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/TransformFunctionTest.java
```diff
@@ -157,6 +203,48 @@ void testDropFields() throws Exception {
...
   }
 
+  @Test
+  void testComputeFields() throws Exception {
+    String userConfig =
+        (""
+                + "{'steps': ["
+                + "    {'type': 'compute-fields', 'fields':["
+                + "        {'name': 'newField1', 'expression' : '5*3', 'type': 'INT32', 'part' : 'key'},"
+                + "        {'name': 'newField2', 'expression' : 'value.valueField1.toUpperCase()', 'type': 'STRING', 'part' : 'key'},"
```
eolivelli, line 213: Generally I don't think it is a good idea to advertise that you can call Java methods.
I am sure that people will do it.

If there is some way to prevent it really would prefer, otherwise we will have probably JDK compatibility issues (if they remove something.....expressions will stop to work)

if we can't stop it, then I would simply do not document it.

keeping it here in the example works for me

eolivelli (approved):
Overall LGTM
I would set the field as "optional" by default, as it is easier to handle for users

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/TransformFunctionTest.java
aymkhalil: it should be possible to disable method invocations. Let me try it along with a utility method as a proof concept that we don't need the first, and we can provide users with the latter...

aymkhalil:
Major updates since the previous patches:
1. Disable bean method invocations (things like `${fieldName.toUpperString()}`)
2. Added sample implementation of custom methods registered under `fn` namespace. Functions for now are String manipulation [`uppercase(), lowercase(), contains()`] and generic `coalesce` function. The way to write an expression for those is `${fn:lowercase('string')}`
3. Made fields optional by default (i.e. union schema with nullable)
4. Remove the `part` field altogether. So instead of:
```JSON
{"type": "compute-fields", "fields": [{"name": "c", "expression" :"'value'", "type" : "STRING", "part" : "value"}]}
```

user would type:
```JSON
{"type": "compute-fields", "fields": [{"name": "value.c", "expression" :"'value'", "type" : "STRING"}]}
```
5. Enable header modification. The only allowed header for now is `destinationTopic` to enable routing as a sample use case.   Example:
```JSON
{"type": "compute-fields", "fields": [{"name": "destinationTopic", "expression" :"destinationTopic == 'targetTopic' ? 'route' : 'dont-route'", "type" : "STRING"}]}
```
notice that for header modifications, the `header.` prefix is not required just like the predicates.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.pulsar.functions.transforms.jstl;
+
+/** Provides convenience methods to use in jstl expression. All functions should be static. */
+public class JstlFunctions {
+
+  public static String uppercase(String input) {
+    return input.toUpperCase();
```
eolivelli, line 22: follow up: handle nulls.
uppercase of null = null
this will make it easier for users to deal with optional fields (and very often every field is optional)

eolivelli (approved):
LGTM

PR #23, opened 2022-10-01
Add config model to leverage jackson mapper

aymkhalil:
This patch adds concrete model that represents step config. It will simplify few things like:
* Avoid parsing the configs manually from JsonNode
* Handling of default values of option fields. 
* Leverage the optional types for config null checks

Note:
* `jackson-datatype-jdk8` is added to support java.util.optional
* Because we have customer error validation, it looks like we are using jackson mapper twice. However the validation logic should be removed at some point when the `networknt` validation limitations are fixed.

### DIFF
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
@@ -17,12 +17,18 @@ package com.datastax.oss.pulsar.functions.transforms;
 
 import com.datastax.oss.pulsar.functions.transforms.jstl.predicate.JstlPredicate;
 import com.datastax.oss.pulsar.functions.transforms.jstl.predicate.StepPredicatePair;
-import com.datastax.oss.pulsar.functions.transforms.jstl.predicate.TransformPredicate;
 import com.datastax.oss.pulsar.functions.transforms.model.ComputeField;
-import com.datastax.oss.pulsar.functions.transforms.model.ComputeFieldType;
+import com.datastax.oss.pulsar.functions.transforms.model.config.CastConfig;
+import com.datastax.oss.pulsar.functions.transforms.model.config.ComputeFieldsConfig;
+import com.datastax.oss.pulsar.functions.transforms.model.config.DropFieldsConfig;
+import com.datastax.oss.pulsar.functions.transforms.model.config.FlattenConfig;
+import com.datastax.oss.pulsar.functions.transforms.model.config.StepConfig;
+import com.datastax.oss.pulsar.functions.transforms.model.config.TransformStepConfig;
+import com.datastax.oss.pulsar.functions.transforms.model.config.UnwrapKeyValueConfig;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import com.networknt.schema.JsonSchema;
 import com.networknt.schema.JsonSchemaFactory;
 import com.networknt.schema.SpecVersion;
@@ -114,6 +120,7 @@ public class TransformFunction
   public void initialize(Context context) {
     Map<String, Object> userConfigMap = context.getUserConfigMap();
     ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
+    mapper.registerModule(new Jdk8Module()); // Adds support for java.util.Optional
     JsonNode jsonNode = mapper.convertValue(userConfigMap, JsonNode.class);
 
     URNFactory urnFactory =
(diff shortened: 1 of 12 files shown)


cbornet:
Actually the models could be generated front the JSON schema spec if we find a good generator.

cbornet:
If we want to go one step further, Jackson has support for polymorphism based on a property (`type` in our case)

aymkhalil:
> If we want to go one step further, Jackson has support for polymorphism based on a property (type in our case)
Is the support your are referring to different to what I'm using in this PR [here](https://github.com/datastax/pulsar-transformations/pull/23/files#diff-d51e38777f0a0275ba618f5da95a492bae2ab74f276357797c1e82dfa8f88a39R24)? I defined an abstract step class and each concrete step will extend it.

aymkhalil:
> Actually the models could be generated from the JSON schema spec if we find a good generator.

I experimented a bit with https://github.com/joelittlejohn/jsonschema2pojo by adding the maven plugin. I was able to get the generated code. There is an issue though with inheritance. ([here](https://github.com/datastax/pulsar-transformations/commit/2190decccec976737ea3068776272aa0f6337979) is the generated models if you wanna have a look)

The configs should extend the Step schema. I tired the example from [here](https://www.jsonschema2pojo.org/): 
```
 {
        "type" : "object",
        "extends" : {
            "$ref" : "flower.json"
        }
    }
```

by using:
```
"$ref": "#/components/schemas/Step"
```
  
However it doesn't work. It seems a separate JSON file is required. Also tried with `javaType` as mentioned [here](https://github.com/joelittlejohn/jsonschema2pojo/issues/402) but didn't work as well.

I think at this stage with few configs a manual model is fine (we also get control over the class names) and automation maybe not be necessary. This is also something that can be iterated on.

I suggest we use the manual model (or drop the PR if not much added value) and invest in code generation later. WDYT @cbornet?

cbornet:
Yes we can keep a manual model. The nice thing with generation would have been to have a single source of truth. Be sure that the schema aligns with the code.

cbornet:
> Is the support your are referring to different from what I'm using in this PR here? I defined an abstract step class and each concrete step will extend it.

No that's exactly it. I had missed that you were using it. That's great!

cbornet:
> The configs should extend the Step schema. 

So jsonschema2pojo has an extension mechanism for polymorphism. I agree that we can look at making it work later. It also seems that the refs are not correctly resolved.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/config/CastConfig.java
```diff
@@ -0,0 +1,28 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Optional;
+import lombok.Getter;
+
+@Getter
+public class CastConfig extends StepConfig {
+  @JsonProperty(value = "schema-type", required = true)
+  private String schemaType;
+
+  private Optional<String> part = Optional.empty();
```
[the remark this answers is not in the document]
aymkhalil: yeah I remember this from another feedback you gave, I do agree that optionals are anti-pattern for business logics classes, but for config models only stuff I found them useful because they communicate clearly that this part of the config is optional with default value equal to null so the person using this config will not forget to handle nulls.

However they don't add huge value in our case (and always good to remove dependency!) - I'll remove it.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/config/DropFieldsConfig.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.List;
+import lombok.Getter;
+
+@Getter
+public class DropFieldsConfig extends StepConfig {
+  @JsonProperty(required = true)
+  private List<String> fields;
+
+  @JsonProperty(required = false)
+  private String part;
```
aymkhalil, line 28: Annotating with @JsonProperty (althought required = false is the defaul) avoid this error
```
[ERROR] Medium: Unwritten field: com.datastax.oss.pulsar.functions.transforms.model.config.DropFieldsConfig.part [com.datastax.oss.pulsar.functions.transforms.model.config.DropFieldsConfig] At DropFieldsConfig.java:[line 27] UWF_UNWRITTEN_FIELD
```

Another we to do it is to add Setter - although jackson deserialization doesn't need it.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
```diff
@@ -284,36 +286,34 @@ public static CastStep newCastFunction(JsonNode node) {
...
+  public static FlattenStep newFlattenFunction(FlattenConfig config) {
     FlattenStep.FlattenStepBuilder builder = FlattenStep.builder();
-    if (node.hasNonNull("part")) {
-      builder.part(node.get("part").asText());
+    if (config.getPart() != null) {
+      builder.part(config.getPart());
     }
-    if (node.hasNonNull("delimiter")) {
-      builder.delimiter(node.get("delimiter").asText());
+    if (config.getDelimiter() != null) {
```
cbornet, line 294: This check is probably not needed ?

cbornet (approved):

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/config/DropFieldsConfig.java
cbornet: That's a spotbug error. Maybe it can deactivated for the Jackson files ?
Or annotate the class with `@Data`. It will make it mutable but is it a big deal ?

cbornet: Or annotate the class with
```java
@Jacksonized
@Builder
@Value
```
to keep it immutable. Jackson will use the builder (https://projectlombok.org/features/experimental/Jacksonized)

eolivelli (approved):
LGTM

Issue #20, opened 2022-09-22
[feat][compute-fields] Support date/time data types 


aymkhalil:
`compute-fields` transform only supports these types today: `[STRING,INT32,INT64,FLOAT,DOUBLE,BOOLEAN]`. The ask is support date related types like: `[Data, Time, Timestamp]`. 

Things to consider:
1) What data format should the user use (fixed vs. allow them to pass the date + format)
2) Should we add custom functions to EL binding to facilitate create those types, or can we use a built in ones?

PR #25, opened 2022-10-04
Support date/time compute fields

aymkhalil:
Fixes: #20 

This patch adds support for the following date/time types:
* `Date`:  Contains date info only, represented as ISO 8601 date like '2022-10-02'
* `Time`: Contains time info only, represented as ISO 8601 time like '23:00:00'
* `OffsetDateTime`: Full date time info,  represented as UTC ISO 3339 like '2011-12-03T10:15:30Z' or '2011-12-03T10:15:30+00:00' or '2011-12-03T10:15:30+02:00'
* Added the following utility date functions:
  * now() return the epoch millis as of now
  * dateadd(input , delta, unit) where
    * input: Is either a Long representing epoch millis or a string representing ISO date like '2011-12-03T10:15:30'
    * delta: As int representing how many units to add or subtract form the input date. 
    * unit: could be one of `[years, months, days, hours, minutes, seconds, millis]`

Implementation note:
* The JSTL evaluator maps the types `Date`, `Time`, `OffsetDateTime` to `java.time.LocalDate`, `java.time.LocalTime`, `java.time.Instant` respectively. 
* In Avro, the types `Date`, `Time`, `OffsetDateTime` are converted to logical times `date`, `time-millis`, timestamp-millis respectively

### DIFF
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
@@ -19,9 +19,9 @@ import static org.apache.pulsar.common.schema.SchemaType.AVRO;
 
 import com.datastax.oss.pulsar.functions.transforms.model.ComputeField;
 import com.datastax.oss.pulsar.functions.transforms.model.ComputeFieldType;
-import java.time.Instant;
 import java.time.LocalDate;
 import java.time.LocalTime;
+import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -202,9 +202,9 @@ public class ComputeFieldStep implements TransformStep {
         LocalTime localTime = (LocalTime) value;
         return (int) (localTime.toNanoOfDay() / 1000000);
       case "timestamp-millis":
-        validateLogicalType(value, schema.getLogicalType(), Instant.class);
-        Instant instant = (Instant) value;
-        return instant.toEpochMilli();
+        validateLogicalType(value, schema.getLogicalType(), OffsetDateTime.class);
+        OffsetDateTime offsetDateTime = (OffsetDateTime) value;
+        return offsetDateTime.toInstant().toEpochMilli();
     }
 
     throw new IllegalArgumentException(
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
@@ -19,10 +19,9 @@ import static org.apache.pulsar.common.schema.SchemaType.AVRO;
 
 import com.datastax.oss.pulsar.functions.transforms.model.ComputeField;
 import com.datastax.oss.pulsar.functions.transforms.model.ComputeFieldType;
+import java.time.Instant;
 import java.time.LocalDate;
-import java.time.LocalDateTime;
 import java.time.LocalTime;
-import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -203,9 +202,9 @@ public class ComputeFieldStep implements TransformStep {
         LocalTime localTime = (LocalTime) value;
         return (int) (localTime.toNanoOfDay() / 1000000);
       case "timestamp-millis":
-        validateLogicalType(value, schema.getLogicalType(), LocalDateTime.class);
-        LocalDateTime localDateTime = (LocalDateTime) value;
-        return localDateTime.toInstant(ZoneOffset.UTC).toEpochMilli();
+        validateLogicalType(value, schema.getLogicalType(), Instant.class);
+        Instant instant = (Instant) value;
+        return instant.toEpochMilli();
     }
 
     throw new IllegalArgumentException(
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
@@ -19,6 +19,10 @@ import static org.apache.pulsar.common.schema.SchemaType.AVRO;
 
 import com.datastax.oss.pulsar.functions.transforms.model.ComputeField;
 import com.datastax.oss.pulsar.functions.transforms.model.ComputeFieldType;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -26,6 +30,8 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import lombok.Builder;
+import org.apache.avro.LogicalType;
+import org.apache.avro.LogicalTypes;
 import org.apache.avro.Schema;
 import org.apache.avro.SchemaBuilder;
 import org.apache.avro.generic.GenericRecord;
@@ -167,10 +173,64 @@ public class ComputeFieldStep implements TransformStep {
     }
     // Add computed fields
     for (ComputeField field : fields) {
-      newRecordBuilder.set(field.getName(), field.getEvaluator().evaluate(context));
-      ;
+      newRecordBuilder.set(
+          field.getName(),
+          getAvroValue(
+              newSchema.getField(field.getName()).schema(),
+              field.getEvaluator().evaluate(context)));
     }
     return newRecordBuilder.build();
+  }
+
+  private Object getAvroValue(Schema schema, Object value) {
+    if (value == null) {
+      return null;
+    }
+
+    LogicalType logicalType = getLogicalType(schema);
+    if (logicalType == null) {
+      return value;
+    }
+
+    // Avro logical type conversion: https://avro.apache.org/docs/1.8.2/spec.html#Logical+Types
+    switch (logicalType.getName()) {
+      case "date":
+        validateLogicalType(value, schema.getLogicalType(), LocalDate.class);
+        LocalDate localDate = (LocalDate) value;
+        return (int) localDate.toEpochDay();
+      case "time-millis":
+        validateLogicalType(value, schema.getLogicalType(), LocalTime.class);
+        LocalTime localTime = (LocalTime) value;
+        return (int) (localTime.toNanoOfDay() / 1000000);
+      case "timestamp-millis":
+        validateLogicalType(value, schema.getLogicalType(), LocalDateTime.class);
+        LocalDateTime localDateTime = (LocalDateTime) value;
+        return localDateTime.toInstant(ZoneOffset.UTC).toEpochMilli();
+    }
+
+    throw new IllegalArgumentException(
+        String.format("Invalid logical type %s for value %s", schema.getLogicalType(), value));
+  }
+
+  private LogicalType getLogicalType(Schema schema) {
+    if (!schema.isUnion()) {
+      return schema.getLogicalType();
+    }
+
+    return schema
+        .getTypes()
+        .stream()
+        .filter(subSchema -> subSchema.getLogicalType() != null)
+        .map(Schema::getLogicalType)
+        .findAny()
+        .orElse(null);
+  }
+
+  void validateLogicalType(Object value, LogicalType logicalType, Class expectedClass) {
+    if (!(value.getClass().equals(expectedClass))) {
+      throw new IllegalArgumentException(
+          String.format("Invalid value %s for logical type %s", value, logicalType));
+    }
   }
 
   private Schema.Field createAvroField(ComputeField field) {
@@ -190,9 +250,12 @@ public class ComputeFieldStep implements TransformStep {
         schemaType = Schema.Type.STRING;
         break;
       case INT32:
+      case DATE:
+      case TIME:
         schemaType = Schema.Type.INT;
         break;
       case INT64:
+      case DATETIME:
         schemaType = Schema.Type.LONG;
         break;
       case FLOAT:
@@ -208,6 +271,21 @@ public class ComputeFieldStep implements TransformStep {
         throw new UnsupportedOperationException("Unsupported compute field type: " + type);
     }
 
-    return fieldTypeToAvroSchemaCache.computeIfAbsent(type, (ignored) -> Schema.create(schemaType));
+    return fieldTypeToAvroSchemaCache.computeIfAbsent(
+        type,
+        key -> {
+          // Handle logical types: https://avro.apache.org/docs/1.10.2/spec.html#Logical+Types
+          Schema schema = Schema.create(schemaType);
+          switch (key) {
+            case DATE:
+              return LogicalTypes.date().addToSchema(schema);
+            case TIME:
+              return LogicalTypes.timeMillis().addToSchema(schema);
+            case DATETIME:
+              return LogicalTypes.timestampMillis().addToSchema(schema);
+            default:
+              return schema;
+          }
+        });
   }
 }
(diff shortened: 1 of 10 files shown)


eolivelli (approved):
LGTM

some follow up steps:
- how does this patch play with the Cassandra Sink, the ElasticSearch Sink (or the other Sinks) ?
- we should add some built-in functions to get the "current date", "current time" and "current datetime", the tricky part is to deal with the timezone, we can have something like now() (uses UTC) and now(timezone) to get the "current date time"
- we should add some built-in function to manipulate these new data types, like adding a offset or computing the difference between two instants

aymkhalil:
> The approach LGTM
> 
> how does this patch play with the Cassandra Sink, the ElasticSearch Sink (or the other Sinks) ?

It works with with ES sink but not with Cassandra Sink (because the logical types that are supported are the CDC custom ones, not standard Avro). I created an [issue](https://github.com/datastax/pulsar-sink/issues/44) to support that.

As far as this patch, it makes sense to stick only to the standard avro types.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/CustomTypeConverter.java
```diff
@@ -17,17 +17,29 @@
...
-public class NullableTypeConverter implements TypeConverter {
+public class CustomTypeConverter implements TypeConverter {
   private static TypeConverterImpl typeConverter = new TypeConverterImpl();
 
   @Override
   public <T> T convert(Object o, Class<T> aClass) throws ELException {
-    return o == null ? null : typeConverter.convert(o, aClass);
+    if (o == null) {
+      return null;
+    } else if ("java.time.LocalDate".equals(aClass.getName())) {
```
eolivelli, line 36: why aren't we using "==" =
`java.time.LocalDate.class == aClass`

it is far more efficient

eolivelli (approved):
LGTM
great stuff

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
```diff
@@ -148,12 +154,66 @@ private GenericRecord computeFields(
...
+      case "date":
+        validateLogicalType(value, schema.getLogicalType(), LocalDate.class);
+        LocalDate localDate = (LocalDate) value;
+        return (int) localDate.toEpochDay();
+      case "time-millis":
+        validateLogicalType(value, schema.getLogicalType(), LocalTime.class);
+        LocalTime localTime = (LocalTime) value;
+        return (int) (localTime.toNanoOfDay() / 1000000);
+      case "timestamp-millis":
+        validateLogicalType(value, schema.getLogicalType(), LocalDateTime.class);
```
cbornet, line 187: Shouldn't it be mapped to a j.u.Instant ?

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstEvaluatorTest.java
```diff
@@ -46,6 +48,47 @@ void testFunctions(String expression, TransformContext context, Object expectedV
...
+    TransformContext primitiveStringContext =
+        new TransformContext(
+            new Utils.TestContext(primitiveStringRecord, new HashMap<>()),
+            primitiveStringRecord.getValue().getNativeObject());
+
+    long expectedNow = Instant.now().toEpochMilli();
+    long actualNow =
+        new JstlEvaluator<Long>("${fn:now()}", long.class).evaluate(primitiveStringContext);
+
+    assertTrue(actualNow >= expectedNow);
```
cbornet, line 67: Maybe we could have the possibility to pass a custom Clock to the class under test.
This way it's easier to control time.

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstEvaluatorTest.java
```diff
@@ -46,6 +48,47 @@ void testFunctions(String expression, TransformContext context, Object expectedV
...
+            AutoConsumeSchema.wrapPrimitiveObject("test-message", SchemaType.STRING, new byte[] {}),
+            "header-key");
+    TransformContext primitiveStringContext =
+        new TransformContext(
+            new Utils.TestContext(primitiveStringRecord, new HashMap<>()),
+            primitiveStringRecord.getValue().getNativeObject());
+
+    long expected = Instant.now().plusSeconds(-3333).toEpochMilli();
+    long actual =
+        new JstlEvaluator<Long>("${fn:dateadd(fn:now(), -3333, 'seconds')}", long.class)
```
cbornet, line 85: I think it would be nice to be able to do pass an ISO8601 duration: https://en.m.wikipedia.org/wiki/ISO_8601#Durations. 
Jackson has good support for deserializing Durations.

cbornet: Jackson may not even be needed. It's probably as simple as Duration.parse() : https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
aymkhalil: This is driven by the type generated from the JSTL evaluator. Now the same question applies (why not use Instant with ejstl?)

We could, I just found it:
1. More consistent with the Time and Date types (because for those the better option is LocalTime & LocalDate as those parse the provided input in ISO format parts out of the box - thinks like `2017-02-10` or `22:30:00`
2. Instant iso format require the user to pass empty Z in the format (like `2017-02-10T22:30:00Z`, but local date time accepts `2017-02-10T22:30:00`. As far as user is concerned, all dates in their avro output will be in reference to epoch
3. LocalDateTime also provides `plus` operation with Units of years and months which as used with the `dateadd` method. Instant only accepts those values: https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#plus-long-java.time.temporal.TemporalUnit-

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstEvaluatorTest.java
aymkhalil: +1, will add in a separate PR: https://github.com/datastax/pulsar-transformations/issues/28

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstEvaluatorTest.java
aymkhalil: JUEL only supports binding of static method, so I have to do it the suboptimal way by introducing a clock on the static class and modify it by tests (as opposed to dependency injection)

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
cbornet: It seems to me that if the datetime is referenced to epoch, Instant is the type to use whereas LocalDatetime is not bound to a timezone (it's the time you see on the clock on the wall without knowing where you are). But probably you transparently assume UTC timezone and the user doesn't see the LocalDatetime type anyway.

aymkhalil: that is fair, let me switch to time instant. And then we can decide separately if we want the user format to be:
`'2007-12-03T10:15:30` vs `2007-12-03T10:15:30.00Z` (the latter is the Instant requirement, empty Z must be included indicating UTC.

Either way, the user is not aware of the underlying date object - it is just easier for us to use the underlying parse method as is without pre-processing

aymkhalil: I don't really know why Instant forces the Z part since it is always strictly UTC!

cbornet: OK I see. What I would suggest is that we support [RFC3339](https://www.rfc-editor.org/rfc/rfc3339#section-5.6) dates. This maps well to j.u.t.OffsetDateTime.
With maybe some tolerance on the input string where if the time offset is not there, we assume UTC. I think Jackson does it (not sure).

cbornet: Also OffsetDateTime has `plusMonths` and `plusYears`

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
```diff
@@ -44,4 +55,55 @@ public static String concat(Object first, Object second) {
...
+
+  private static long dateadd(long epochMillis, long delta, String unit) {
+    Instant instant = Instant.ofEpochMilli(epochMillis);
+    LocalDateTime localDateTime = LocalDateTime.ofInstant(instant, UTC);
+    return dateadd(localDateTime, delta, unit);
+  }
+
+  private static long dateadd(LocalDateTime localDateTime, long delta, String unit) {
+    ChronoUnit chronoUnit = dateAddUnits.get(unit);
+    if (chronoUnit == null) {
```
cbornet, line 102: We could also accept if `ChronoUnit.valueOf(unit)` returns something ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
```diff
@@ -44,4 +55,55 @@ public static String concat(Object first, Object second) {
...
+  }
+
+  private static long dateadd(long epochMillis, long delta, String unit) {
+    Instant instant = Instant.ofEpochMilli(epochMillis);
+    LocalDateTime localDateTime = LocalDateTime.ofInstant(instant, UTC);
+    return dateadd(localDateTime, delta, unit);
+  }
+
+  private static long dateadd(LocalDateTime localDateTime, long delta, String unit) {
+    ChronoUnit chronoUnit = dateAddUnits.get(unit);
```
cbornet, line 101: We could use ChronoUnit.valueOf() and remove the dateAddUnits map ?
```java
  public static long dateadd(Object input, long delta, String unit) {
    if (input instanceof String) {
      return dateadd((String) input, delta, unit);
    } else if (input instanceof Long) {
      return dateadd((long) input, delta, unit);
    }

        "Invalid input type: "
            + input.getClass().getSimpleName()
            + ". Should either be an RFC3339 datetime string like '2007-12-01T12:30:00Z' or epoch millis");
  }

  private static long dateadd(String isoDateTime, long delta, String unit) {
    OffsetDateTime offsetDateTime = OffsetDateTime.parse(isoDateTime);
    return dateadd(offsetDateTime, delta, unit);
  }

  private static long dateadd(long epochMillis, long delta, String unit) {
    Instant instant = Instant.ofEpochMilli(epochMillis);
    OffsetDateTime offsetDateTime = OffsetDateTime.ofInstant(instant, UTC);
    return dateadd(offsetDateTime, delta, unit);
  }

  private static long dateadd(OffsetDateTime offsetDateTime, long delta, String unit) {
    try {
      ChronoUnit chronoUnit = ChronoUnit.valueOf(unit.toUpperCase());
      return offsetDateTime.plus(delta, chronoUnit).toInstant().toEpochMilli();
    } catch (IllegalArgumentException e) {
          "Invalid unit: "
              + unit
              + ". Should be one of "
              + Arrays.stream(ChronoUnit.values())
                  .map(s -> s.name().toLowerCase())
                  .collect(Collectors.toList()),
          e);
    }
  }
```

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/jstl/CustomTypeConverterTest.java
```diff
@@ -19,29 +19,42 @@
...
+    CustomTypeConverter converter = new CustomTypeConverter();
     assertNull(converter.convert(null, String.class));
     assertNull(converter.convert(null, int.class));
     assertNull(converter.convert(null, long.class));
     assertNull(converter.convert(null, float.class));
     assertNull(converter.convert(null, double.class));
     assertNull(converter.convert(null, boolean.class));
+    assertNull(converter.convert(null, LocalDate.class));
+    assertNull(converter.convert(null, LocalTime.class));
+    assertNull(converter.convert(null, LocalDateTime.class));
```
cbornet, line 40: Still LocalDateTime ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
```diff
@@ -44,4 +55,55 @@ public static String concat(Object first, Object second) {
   public static Object coalesce(Object value, Object valueIfNull) {
     return value == null ? valueIfNull : value;
   }
+
+  public static long now() {
+    return clock.millis();
+  }
+
+  private static final ZoneId UTC = ZoneId.of("UTC");
```
cbornet, line 63: Already exists as ZoneOffset.UTC

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
aymkhalil: The unit does not map one to one with the Chrono units (it is a subset):
`[nanos, micros, millis, seconds, minutes, hours, half_days, days, weeks, months, years, decades, centuries, millennia, eras, forever]` vs `years, months, days,hours,minutes,millis` - also just preder to have them decoupled in case we use different names (day vs days)

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
aymkhalil: +1 to ISO 3330. to make documentation easier, we can stick to its format and refer user to the RFC. (for UTC, they could write either 2022-07-02T01:02:00Z OR 2022-07-02T01:02:00+00). If the provide millis (which I think will be common case) we assume epoch millis. Another common use case that won't include the string is fn:now()).

we can always add more permissive format (like 2022-07-02T01:02:00) without breaking existing users...

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
cbornet: You're right. It's better to keep it decoupled.
Maybe use a switch instead of a Map ? (no strong opinion here)
If we keep the Map, we should initialize it with Map.of() (min JDK is 11) instead of a static initializer.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
```diff
@@ -167,12 +172,66 @@ private GenericRecord computeFields(
...
+  private LogicalType getLogicalType(Schema schema) {
+    if (!schema.isUnion()) {
+      return schema.getLogicalType();
+    }
+
+    return schema
+        .getTypes()
+        .stream()
+        .filter(subSchema -> subSchema.getLogicalType() != null)
+        .map(Schema::getLogicalType)
```
cbornet, line 223: The filter and map can be switched. And filter is then just `Objects::nonnull`

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
aymkhalil: +1. Switch fits better, I wanted to cache them for some reason but there is no need to.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeFieldStep.java
aymkhalil: +1

PR #29, opened 2022-10-05
Add support for header properties compute

### LINKED ISSUE #27: [feat][compute] Support computing custom KV props
Enable users to add KV custom props to the message header props Map

aymkhalil:
(Fixes #27)

This patch adds support for computing custom fields on the properties Key/Value headers. 

**Implementation note:**
The behavior of the transform context is change from overwriting the current record props to merging - only computed props that have matching keys will overwrite existing ones.


eolivelli (approved):
LGTM

PR #30, opened 2022-10-14
Rename ComputeFields step to Compute

aymkhalil:
Rename `ComputeFields` step to `Compute` as the later is more general. Note that compute can be applied to key fields, values fields, and selected header fields like `destinationTopic` and `properties` KV map.


eolivelli (approved):
LGTM

PR #36, opened 2022-10-26
Cache flattened schemas

aymkhalil:
Optimize flatten step by:
1. Cache the schemas of flattened parts (key or value)
2. Decouple flattened field value evaluation from schema evaluation. This would enable caching the schema once, and use it to calculate field values at each transform invocation,
3. Add simple JMH benchmarking to establish a baseline of improvements

Sample benchmark before caching (1000 nested levels):
```
Benchmark                       (nestedLevels)  Mode  Cnt   Score    Error  Units
FlattenStepBenchmark.doFlatten              10  avgt    5  ≈ 10⁻⁴            s/op
FlattenStepBenchmark.doFlatten             100  avgt    5   0.002 ±  0.001   s/op
FlattenStepBenchmark.doFlatten            1000  avgt    5   0.147 ±  0.010   s/op

```

and after (notice it is actually worse for the 1000 levels run):
```
FlattenStepBenchmark.doFlatten              10  avgt    5  ≈ 10⁻⁴            s/op
FlattenStepBenchmark.doFlatten             100  avgt    5   0.001 ±  0.001   s/op
FlattenStepBenchmark.doFlatten            1000  avgt    5   0.348 ±  0.003   s/op
```

### DIFF
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/FlattenStep.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/FlattenStep.java
@@ -16,10 +16,15 @@
 package com.datastax.oss.pulsar.functions.transforms;
 
 import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
-import javax.annotation.Nullable;
+
 import lombok.Builder;
+import lombok.RequiredArgsConstructor;
 import lombok.Value;
 import org.apache.avro.SchemaBuilder;
 import org.apache.avro.generic.GenericData;
@@ -40,6 +45,10 @@ public class FlattenStep implements TransformStep {
 
   @Builder.Default private final String delimiter = DEFAULT_DELIMITER;
   private final String part;
+  private final Map<org.apache.avro.Schema, FlattenedSchema> keySchemaCache =
+      new ConcurrentHashMap<>();
+  private final Map<org.apache.avro.Schema, FlattenedSchema> valueSchemaCache =
+      new ConcurrentHashMap<>();
 
   @Override
   public void process(TransformContext transformContext) throws Exception {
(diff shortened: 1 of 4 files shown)


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/FlattenStep.java
```diff
@@ -78,30 +87,53 @@ void validateAvro(Schema<?> schema) {
...
     return newRecord;
   }
 
-  private List<FieldValuePair> buildFlattenedFields(GenericRecord record) {
-    List<FieldValuePair> flattenedFields = new ArrayList<>();
-    org.apache.avro.Schema originalSchema = record.getSchema();
-    for (org.apache.avro.Schema.Field field : originalSchema.getFields()) {
-      flattenedFields.addAll(
-          flattenField(record, record.getSchema(), field, field.schema().isNullable(), ""));
+  private Object getFlattenedFieldValue(
```
aymkhalil, line 108: this can be further optimized, although maybe more complex than necessary.

aymkhalil: this improved ~10x with `parallelStream` on the loop calling this method - 118ms avg -> 17ms avg. 

```
    modified.getFieldMap().entrySet().parallelStream().forEach(e -> newRecord.put(e.getKey().name(), getFlattenedFieldValue(e.getValue(), record)));

```

after all, practical performance improvements (if needed) maybe a result of run some parts of code in parallel, rather than caching.

eolivelli (approved):
+1

eolivelli:
is it still a draft ?

aymkhalil:
> is it still a draft ?

It is. I'm was looking for feedback on performance improvement. For now I think the added complexity is not justified. Here is a baseline benchmark (without caching):
```
FlattenStepBenchmark.doFlatten              10  avgt    5  ≈ 10⁻⁴            s/op
FlattenStepBenchmark.doFlatten             100  avgt    5   0.002 ±  0.001   s/op
FlattenStepBenchmark.doFlatten            1000  avgt    5   0.147 ±  0.010   s/op
```
I'm thinking to close this PR for now and reconsider if there is a need. Also an event with 1000 nested levels is a big stretch!

Issue #38, opened 2023-01-06
[improvement][devex] Add the ability to compute primitive schema type


aymkhalil:
Today the compute step only allows modifications of fields in an avro schema or a select set of headers (like destination topics). Attempt to modify the `value` for example for primitive types will fail with:
```
Invalid compute field name: value. It should be prefixed with 'key.' or 'value.' or 'properties.' or be one of [destinationTopic, messageKey]
``` 

The idea is to enable users to perform simple stuff like schema conversion or concatenating string types like:
```
{
  "steps": [
    {
      "type": "compute", 
      "fields": [ 
        {
          "name": "value", // e.g this could be a STRING type with value 1.3   
          "expression" : "value" 
          "type": "DOUBLE" // effectively does the casting
        }
      ]
    }
  ]
}
``` 

or 

```
{
  "steps": [
    {
      "type": "compute", 
      "fields": [ 
        {
          "name": "value", 
          "expression" : "fn:(value, '!'") 
          "type": "STRING"
        }
      ]
    }
  ]
}
```

cbornet:
Fixed by #39

PR #39, opened 2023-01-07
Support compute step on primitive schema types

aymkhalil:
Addresses: #38 

This patch adds support for computing primitive schema types. Implementation notes:
1. Add a new validation on the compute steps config to ensure uniqueness. In case of overwriting a primitive type value, the compute field will be `value` and the validation will guarantee a single compute name `value`. An alternative approach would be to pick the last one (could be confusing to the user) or cascade the computation (adds complexity, user can always cascade if needed by creating separate steps).
2. Validation is one is custom, ideally it should be in the schema validator but there is no direct way to ensure uniqness based on a key inside array of objects - here are the uniqness use cases implemented for [uniqueItems](https://github.com/networknt/json-schema-validator/blob/416572e159d5d13bffa43a04d578fd7d384ef776/src/test/resources/draft4/uniqueItems.json)
3. The select mapping between compute fields types and generic primitive schema is (the date/time types are the best mapping I could come up in light of Pulsar [schema types](https://pulsar.apache.org/docs/2.10.x/schema-understand/#primitive-type):
```
      ComputeFieldType.STRING -> org.apache.pulsar.client.api.Schema.STRING;
      ComputeFieldType.INT32 -> org.apache.pulsar.client.api.Schema.INT32; 
      ComputeFieldType.INT64 -> org.apache.pulsar.client.api.Schema.INT64;
      ComputeFieldType.FLOAT -> org.apache.pulsar.client.api.Schema.FLOAT;
      ComputeFieldType.DOUBLE ->  org.apache.pulsar.client.api.Schema.DOUBLE;
      ComputeFieldType.BOOLEAN -> org.apache.pulsar.client.api.Schema.BOOL;
      ComputeFieldType.DATE -> org.apache.pulsar.client.api.Schema.DATE;
      ComputeFieldType.TIME ->  org.apache.pulsar.client.api.Schema.TIME;
      ComputeFieldType.DATETIME -> org.apache.pulsar.client.api.Schema.TIMESTAMP;
```


eolivelli (approved):
LGTM

great stuff

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeStep.java
```diff
@@ -288,4 +303,41 @@ private Schema getAvroSchema(ComputeFieldType type) {
...
+      case BOOLEAN:
+        schema = org.apache.pulsar.client.api.Schema.BOOL;
+        break;
+      case DATE:
+        schema = org.apache.pulsar.client.api.Schema.DATE;
+        break;
+      case TIME:
+        schema = org.apache.pulsar.client.api.Schema.TIME;
+        break;
+      case DATETIME:
```
cbornet, line 334: I think we should have the same names.
So `ComputeFieldType.TIMESTAMP` -> `Schema.TIMESTAMP`.
And support the other primitive types `INSTANT`, `LOCAL_DATE_TIME`, ...

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/ComputeStepTest.java
```diff
@@ -280,6 +282,39 @@ void testKeyValueAvro() throws Exception {
     assertEquals(messageSchema.getKeyValueEncodingType(), KeyValueEncodingType.SEPARATED);
   }
 
+  @Test(dataProvider = "primitiveTypeComputeProvider")
+  void testPrimitiveSchemaTypes(
+      Object oldValue,
+      Schema oldSchema,
+      SchemaType oldType,
```
cbornet, line 289: We can get the type from the schema

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/ComputeStepTest.java
```diff
@@ -280,6 +282,39 @@ void testKeyValueAvro() throws Exception {
...
+      String expression,
+      ComputeFieldType newSchema,
+      Object expectedValue,
+      Schema expectedSchema)
+      throws Exception {
+    Record<GenericObject> record =
+        new Utils.TestRecord<>(
+            oldSchema,
+            AutoConsumeSchema.wrapPrimitiveObject(oldValue, oldType, new byte[] {}),
```
cbornet, line 298: `AutoConsumeSchema.wrapPrimitiveObject(oldValue, oldSchema.getSchemaInfo().getType(), new byte[] {}),`

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/ComputeStepTest.java
```diff
@@ -280,6 +282,39 @@ void testKeyValueAvro() throws Exception {
     assertEquals(messageSchema.getKeyValueEncodingType(), KeyValueEncodingType.SEPARATED);
   }
 
+  @Test(dataProvider = "primitiveTypeComputeProvider")
+  void testPrimitiveSchemaTypes(
+      Object oldValue,
+      Schema oldSchema,
```
cbornet, line 288: Schema<?> oldSchema,

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/ComputeStepTest.java
```diff
@@ -280,6 +282,39 @@ void testKeyValueAvro() throws Exception {
...
+      SchemaType oldType,
+      String expression,
+      ComputeFieldType newSchema,
+      Object expectedValue,
+      Schema expectedSchema)
```
cbornet, line 293: `      Schema<?> expectedSchema)
`

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/ComputeStepTest.java
```diff
@@ -280,6 +282,39 @@ void testKeyValueAvro() throws Exception {
...
+    Record<GenericObject> record =
+        new Utils.TestRecord<>(
+            oldSchema,
+            AutoConsumeSchema.wrapPrimitiveObject(oldValue, oldType, new byte[] {}),
+            "test-key");
+
+    ComputeStep step =
+        ComputeStep.builder()
+            .fields(
+                Arrays.asList(
```
cbornet, line 304: `Collections.singletonList(`

cbornet (changes requested):

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/ComputeStepTest.java
```diff
@@ -462,6 +497,181 @@ public static Object[] destinationTopicProvider() {
...
+   * @return {"old value", "old schema", "old type", "expression", "new schema", "expected value",
+   *     "expectedSchema"}
+   */
+  @DataProvider(name = "primitiveTypeComputeProvider")
+  public static Object[][] primitiveTypeComputeProvider() {
+    return new Object[][] {
+      {
+        "oldValue",
+        Schema.STRING,
+        SchemaType.STRING,
```
cbornet, line 510: It seems these values don't really matter.
Maybe we could set `"'value'"` for the expression instead ?

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/ComputeStepTest.java
```diff
@@ -462,6 +497,181 @@ public static Object[] destinationTopicProvider() {
...
+      },
+      {"true", Schema.BOOL, SchemaType.BOOLEAN, "1", ComputeFieldType.DOUBLE, 1D, Schema.DOUBLE},
+      {
+        "2007-01-02",
+        Schema.DATE,
+        SchemaType.DATE,
+        "'2008-02-07'",
+        ComputeFieldType.DATE,
+        LocalDate.parse("2008-02-07"),
+        Schema.DATE
```
cbornet, line 616: This schema and value are not compatible.
Schema should be `Schema.LOCAL_DATE` or value should be a `java.util.Date`

cbornet:
There's a small mismatch between the AVRO field types and the Pulsar schema ones.
Currently, for AVRO `ComputeFieldType.DATE --> LocalDate` but for Pulsar Schema, it should be `ComputeFieldType.DATE --> j.u.Date`.
I think we should have the latter and `ComputeFieldType.LOCAL_DATE --> LocalDate` for both AVRO and Pulsar.
And for AVRO, either not support `ComputeFieldType.DATE` or have `ComputeFieldType.DATE --> OffsetDateTime`

aymkhalil:
> Currently, for AVRO ComputeFieldType.DATE --> LocalDate

So there is no 1:1 mapping between Java dates types and AVRO types. As far as user is concerned they are getting an `int` of "e number of days from the unix epoch" as per AVROs date logical type.

This is calculated like so:
```
(int) localDate.toEpochDay();
```

> And for AVRO, either not support ComputeFieldType.DATE or have ComputeFieldType.DATE --> OffsetDateTime

Not supporting DATE seems unnecessarily limiting  the user experience, and `OffsetDateTime` will force the user to have the Time portion in their String or whatever they are converting from, so Date seems the more natural fit. What did I miss?

I agree on the Pulsar schema front, there are some mismatches gaps I'll try to cover...

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/ComputeStepTest.java
aymkhalil: let me resolve the discrepancies in Pulsar API - they directly map to the java types and it makes sense for them to match. I'll also check the exhaustive list of schemas and try to add them (like byes) if they don't blow up the scope of this patch...

aymkhalil: Also `j.u.Date` is deprecated so implementation wise we're using the `LocalDate` - does the Java type returned by the API has to strictly map to the Pulsar Schema API? 

I'm more more thinking now that the Compute field types should actually be our own (not 1:1 mapping with Pulsar API) so we have more control over our implementation and can always document the behavior for users...

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeStep.java
aymkhalil: ComputeFieldType doesn't have a 1TIMESTAMP1. The convention used so far (up for feedback, review) is `TIME` `DATE`, `DATETIME`.  These are meant as logical abstraction on top of the Avro schema, in Avro, the underlying java client will resolve to INT/LONG, but in Pulsar API, the schema types are directly coupled with the Java types and hence the confusion. In principle the user can represent any RFC3339 today with the available compute types, but lets have a discussion about what makes sense for the primitive types use cases...

If we don't go with 1:1 mapping for primitive type, we can still add missing types like `BYTE` but `BYTEBUFFER` for example won't be useful.

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/ComputeStepTest.java
cbornet: > does the Java type returned by the API has to strictly map to the Pulsar Schema API

Yes, Pulsar Schema.DATE can only encode/decode j.u.Date.
I guess we could decide to not support older Java types in `compute`. And only have them in `cast`.
My concern is that from a user perspective there will be a doubt as to whether DATE maps to Date or LocalDate.

aymkhalil: Makes sense. I'll stick to `java.time.*`.

aymkhalil: Previously, we chose `j.t.OffsetDateTime` over `j.t.LocalDateTime` as a better representation of `RFC3339` dates (e.g. it doesn't parse if there is `Z ` in the end). The pulsar schema API thought doesn't have a Schema.OffsetDateTime though. I wonder if in this case it is Ok to return `j.t.OffsetDateTime` or just switch to `j.t.LocalDateTime`

aymkhalil: So I kept `OffsetDateTime` to capture input, and switched output to `Instant` - it seems the only schema time available in the public API today that can represent full RFC3339 date/time.

cbornet:
I mean for AVRO, use `ComputeFieldType.LOCAL_DATE` instead of `ComputeFieldType.DATE`.
To avoid confusion with `SchemaType.DATE`

aymkhalil:
In the latest patch:
* Added support for primitive key type in a key value schema
* Deprecated `DATETIME`
* Added the exhaustive list of Pulsar schema type (INT8, INT16, LOCAL_TIME, LOCAL_DATE, LOCAL_DATE_TIME, INSTANT, TIMESTAMP) 
* Supported that new added type in AVRO 

@cbornet PTAL.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeStep.java
```diff
@@ -249,13 +321,20 @@ private Schema getAvroSchema(ComputeFieldType type) {
       case STRING:
         schemaType = Schema.Type.STRING;
         break;
+      case INT8:
```
aymkhalil, line 324: This could be a FIXED type in Avro. Using INT for simplicity but let me know if it is better to change it. 

If I don't know about the pulsar types, then by default expectation for users to use Avro's INT or LONG.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeStep.java
```diff
@@ -215,23 +244,43 @@ private Object getAvroValue(Schema schema, Object value) {
...
     }
 
     throw new IllegalArgumentException(
         String.format("Invalid logical type %s for value %s", schema.getLogicalType(), value));
   }
 
+  private Object getAvroTimeTimestampMillis(Object value, LogicalType logicalType) {
+    validateLogicalType(value, logicalType, Instant.class, Timestamp.class, LocalDateTime.class);
+    final Instant instant;
+    if (value instanceof Timestamp) {
```
cbornet, line 261: Could we use the CustomTypeConverter ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeStep.java
cbornet: I think it's OK to map to INT atm.
At least it's not semantically wrong. We can change later.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeStep.java
```diff
@@ -195,23 +244,43 @@ private Object getAvroValue(Schema schema, Object value) {
...
+      instant = (Instant) value;
+    }
+    return instant.toEpochMilli();
+  }
+
+  private Object getAvroData(Object value, LogicalType logicalType) {
+    validateLogicalType(value, logicalType, Date.class, LocalDate.class);
+    return (value instanceof LocalDate)
+        ? (int) ((LocalDate) value).toEpochDay()
+        : (int) (((Date) value).getTime() / TimeUnit.DAYS.toMillis(1));
```
cbornet, line 275: Need to extract `TimeUnit.DAYS.toMillis(1)` to a constant. Otherwise it's computed each time.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeStep.java
cbornet: Also AVRO int and long are encoded using variable zig-zag encoding so it's not even sure that you gain bytes by using a fixed type. It depends on the application.

cbornet:
Awesome work @aymkhalil !

PR #40, opened 2023-01-10
Add more conversions to cast

cbornet:


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/CastStep.java
```diff
@@ -21,25 +21,101 @@
...
+        && transformContext.getValueSchema().getSchemaInfo().getType() != valueSchemaType) {
+      Object value = convertValue(transformContext.getValueObject(), valueSchemaType);
+      transformContext.setValueSchema(toSchema(valueSchemaType));
+      transformContext.setValueObject(value);
+    }
+  }
+
+  private Object convertValue(Object originalValue, SchemaType schemaType) {
+    switch (schemaType) {
+      case BYTES:
```
aymkhalil, line 46: There is `BYTEBUFFER` as well if you are targeting the exhaustive list of schemas.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/CastStep.java
```diff
@@ -21,25 +21,101 @@
...
+        return ConverterUtil.toLocalDateTime(originalValue);
+      default:
+        throw new IllegalStateException("Unexpected value: " + schemaType);
     }
-    if (valueSchemaType == SchemaType.STRING) {
-      transformContext.setValueSchema(Schema.STRING);
-      transformContext.setValueObject(transformContext.getValueObject().toString());
+  }
+
+  private Schema<?> toSchema(SchemaType schemaType) {
```
aymkhalil, line 83: I was thinking if those are worth caching but the seem defined as fields in the interface (which translates to static final). Just wondering what you think?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ConverterUtil.java
```diff
@@ -0,0 +1,385 @@
...
+    if (o instanceof LocalDate) {
+      return (LocalDate) o;
+    }
+    if (o instanceof LocalDateTime) {
+      return ((LocalDateTime) o).toLocalDate();
+    }
+    if (o instanceof byte[]) {
+      return Schema.LOCAL_DATE.decode((byte[]) o);
+    }
+    if (o instanceof CharSequence) {
```
aymkhalil, line 304: should this be `o instanceof String`?

aymkhalil (approved):
Great stuff, I added minor comments.

aymkhalil (changes requested):
Forgot to mention, the [config-schema](https://github.com/datastax/pulsar-transformations/blob/b69dd104987d971069b4efca59021195e6aa239a/pulsar-transformations/src/main/resources/config-schema.yaml#L82) model needs an update

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ConverterUtil.java
```diff
@@ -0,0 +1,385 @@
...
+    if (o instanceof Long) {
+      return Instant.ofEpochMilli(((Number) o).longValue());
+    }
+    if (o instanceof Double) {
+      long seconds = (long) ((double) o / 1000);
+      long nanos = Math.round(((double) o - seconds * 1000) * 1_000_000);
+      return Instant.ofEpochSecond(seconds, nanos);
+    }
+    if (o instanceof CharSequence) {
+      return DateTimeFormatter.ISO_DATE_TIME.parse((CharSequence) o, Instant::from);
```
aymkhalil, line 363: I just checked checked how OffsetDateTime works and it actually uses [ISO_OFFSET_DATE_TIME](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_OFFSET_DATE_TIME) . There are nuances between the two that I don't fully understand but we can agree on to use it consistently between the cast and compute steps.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/CastStep.java
cbornet: Yes, primitive Schemas are static so they don't need caching.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/CastStep.java
cbornet: Here it's a `SchemaType` for which there's only `SchemaType.BYTES`

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ConverterUtil.java
aymkhalil: does the same apply to L263?

cbornet: indeed

cbornet: done

cbornet:
> Forgot to mention, the [config-schema](https://github.com/datastax/pulsar-transformations/blob/b69dd104987d971069b4efca59021195e6aa239a/pulsar-transformations/src/main/resources/config-schema.yaml#L82) model needs an update

done
I also updated the README. PTAL.

PR #44, opened 2023-01-13
Add fn:replace function

cbornet:
Most string operations can be done with a good regex. 
Use Chat-GPT to generate that regex for you 😄 

Eg to do `substringBefore("-")` <--> `fn:replace(value, '.*-', '')`

### DIFF
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
@@ -53,6 +53,12 @@ public class JstlFunctions {
     return value == null ? valueIfNull : value;
   }
 
+  public static String replace(Object input, Object regex, Object replacement) {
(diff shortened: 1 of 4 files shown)


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
```diff
@@ -53,6 +53,12 @@ public static Object coalesce(Object value, Object valueIfNull) {
     return value == null ? valueIfNull : value;
   }
 
+  public static String replace(Object input, Object regex, Object replacement) {
```
aymkhalil, line 56: Existing methods doesn't fail if any arg is null. Would be sense here to return `input` if either `regex` or `replacement` is null?

cbornet: done

aymkhalil (approved):

PR #45, opened 2023-01-13
Use tomcat-embeded-el instead of JUEL

aymkhalil:
Today we take dependency on [JUEL](https://github.com/beckchr/juel) to evaluate EL expression in the `Compute` transform and the `Predicate` feature. JUEL is not maintained any more and support an older version of EL (2.0).

In this path, an attempt to replace JUEL with the more active [tomcat-embeded-el](https://mvnrepository.com/artifact/org.apache.tomcat.embed/tomcat-embed-el) is made. Here are few notes:
* Performance of both implementations is very similar using a simple benchmark issue. Nothing conerning.
* Feature wise, I couldn't find an easy way to disable method invocations (which we did for security reasons, here is how how the [feature](https://github.com/beckchr/juel/blob/1a8fb366b7349ddae81f806dee1ab2de11b672c7/modules/impl/src/main/java/de/odysseus/el/ExpressionFactoryImpl.java#L115) can be toggled in JUEL. For now, method invocation tests are disabled.
* String concat is now supported via `+=` operators. Example: `${value += '-suffux'}`
* Moderate code refactoring is made to accommodate `CustomerTypeConverter` - because in `tomcat-embeded-el` the conversion (added via ELResover) is custom aware. And since `CustomerTypeConverter`  is common code with the `CastStep` which doesn't use EL, some adaptation is required.

### DIFF
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/CustomTypeConverter.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/CustomTypeConverter.java
@@ -56,6 +56,7 @@ public class CustomTypeConverter extends TypeConverter {
     return null;
   }
 
+  @Override
   protected Double coerceToDouble(Object value) {
     if (value instanceof LocalTime) {
       return (double) ((LocalTime) value).toNanoOfDay() / 1_000_000;
@@ -80,6 +81,7 @@ public class CustomTypeConverter extends TypeConverter {
     return null;
   }
 
+  @Override
   protected Float coerceToFloat(Object value) {
     if (value instanceof byte[]) {
       return Schema.FLOAT.decode((byte[]) value);
@@ -87,6 +89,7 @@ public class CustomTypeConverter extends TypeConverter {
     return null;
   }
 
+  @Override
   protected Long coerceToLong(Object value) {
     if (value instanceof LocalTime) {
       return ((LocalTime) value).toNanoOfDay() / 1_000_000;
@@ -106,6 +109,7 @@ public class CustomTypeConverter extends TypeConverter {
     return null;
   }
 
+  @Override
   protected Integer coerceToInteger(Object value) {
     if (value instanceof LocalTime) {
       return (int) (((LocalTime) value).toNanoOfDay() / 1_000_000);
@@ -119,6 +123,7 @@ public class CustomTypeConverter extends TypeConverter {
     return null;
   }
 
+  @Override
   protected Short coerceToShort(Object value) {
     if (value instanceof byte[]) {
       return Schema.INT16.decode((byte[]) value);
@@ -126,6 +131,7 @@ public class CustomTypeConverter extends TypeConverter {
     return null;
   }
 
+  @Override
   protected Byte coerceToByte(Object value) {
     if (value instanceof byte[]) {
       return Schema.INT8.decode((byte[]) value);
@@ -133,6 +139,7 @@ public class CustomTypeConverter extends TypeConverter {
     return null;
   }
 
+  @Override
   protected String coerceToString(Object value) {
     if (value instanceof Time) {
       return DateTimeFormatter.ISO_LOCAL_TIME.format(((Time) value).toLocalTime());
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/CustomTypeConverter.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/CustomTypeConverter.java
@@ -15,8 +15,10 @@
  */
 package com.datastax.oss.pulsar.functions.transforms.jstl;
 
-import de.odysseus.el.misc.LocalMessages;
-import de.odysseus.el.misc.TypeConverterImpl;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import jakarta.el.ELContext;
+import jakarta.el.ELException;
+import jakarta.el.TypeConverter;
 import java.sql.Time;
 import java.sql.Timestamp;
 import java.time.Instant;
@@ -29,15 +31,16 @@ import java.time.format.DateTimeFormatter;
 import java.time.temporal.TemporalAccessor;
 import java.util.Date;
 import java.util.Map;
-import javax.el.ELException;
 import org.apache.avro.util.Utf8;
+import org.apache.el.lang.ELSupport;
+import org.apache.el.util.MessageFactory;
 import org.apache.pulsar.client.api.Schema;
 
 /**
  * Overrides the default TypeConverter coerce to support null values & non-EL coercions (e.g.
  * date/time types) schemas.
  */
-public class CustomTypeConverter extends TypeConverterImpl {
+public class CustomTypeConverter extends TypeConverter {
 
   public static final CustomTypeConverter INSTANCE = new CustomTypeConverter();
 
@@ -45,15 +48,14 @@ public class CustomTypeConverter extends TypeConverterImpl {
     super();
   }
 
-  @Override
+  @SuppressFBWarnings("NP_BOOLEAN_RETURN_NULL")
   protected Boolean coerceToBoolean(Object value) {
     if (value instanceof byte[]) {
       return Schema.BOOL.decode((byte[]) value);
     }
-    return super.coerceToBoolean(value);
+    return null;
   }
 
-  @Override
   protected Double coerceToDouble(Object value) {
     if (value instanceof LocalTime) {
       return (double) ((LocalTime) value).toNanoOfDay() / 1_000_000;
@@ -75,18 +77,16 @@ public class CustomTypeConverter extends TypeConverterImpl {
       Instant instant = coerceToInstant(value);
       return (double) instant.getEpochSecond() * 1000 + (double) instant.getNano() / 1_000_000;
     }
-    return super.coerceToDouble(value);
+    return null;
   }
 
-  @Override
   protected Float coerceToFloat(Object value) {
     if (value instanceof byte[]) {
       return Schema.FLOAT.decode((byte[]) value);
     }
-    return super.coerceToFloat(value);
+    return null;
   }
 
-  @Override
   protected Long coerceToLong(Object value) {
     if (value instanceof LocalTime) {
       return ((LocalTime) value).toNanoOfDay() / 1_000_000;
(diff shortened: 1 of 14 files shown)


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/CustomTypeConverter.java
```diff
@@ -394,4 +413,25 @@ protected Instant coerceToInstant(Object value) {
     }
     throw new IllegalArgumentException("Cannot convert type " + value.getClass() + " to Instant");
   }
+
+  public <T> T convert(Object value, Class<T> type) throws ELException {
```
aymkhalil, line 417: There are few options here:
1) Use two different impls, one for compute and one for cast and the common conversion can go in an abstract method
2) Stick to the convertToType interface method and have it deal with null context.

eolivelli (approved):
Lgtm


Please bump the major version number, this is a core change

If users have problems we can maintain two release branches

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/CustomTypeConverter.java
```diff
@@ -28,31 +30,31 @@
...
   }
 
-  @Override
+  @SuppressFBWarnings("NP_BOOLEAN_RETURN_NULL")
   protected Boolean coerceToBoolean(Object value) {
     if (value instanceof byte[]) {
       return Schema.BOOL.decode((byte[]) value);
     }
-    return super.coerceToBoolean(value);
+    return null;
```
cbornet, line 55: I'd rather call `ELSupport.coerceToBoolean` than return null

eolivelli: what type are we expecting here ?
only byte[] ?

aymkhalil: > I'd rather call ELSupport.coerceToBoolean than return null

the issue with that is `tomcat-embeded-el` expects `null` and `context.setPropertyResolved(false);` in order to proceed with other resolvers. So I was thinking to set all `coerceTo*` methods to private, and have a single entry point to this class which is the `convert` methods. It will handle two cases:
1) When used with `compute`, it sets `context.setPropertyResolved(false);` and propagates the null
2) When without context, like in `cast`, it returns `ELSupport.coerceToType`

aymkhalil: > what type are we expecting here ?
only byte[] ?

anything that reasonably maps to boolean (mainly string `"true|false"` or literal `true|false`. This will be taken care of the built in converter (in JUEL or tomcat embedded). Any custom case (like byte[]) we handle our selves like in this example, via the `PULSAR.BOOL` schema decoder...

cbornet: I think it's OK to call ELSupport.coerceToBoolean. It will resolve the property which is exactly what we want. It doesn't matter if it's done now or later.

aymkhalil: If we do that, we'll basically overwrite the underling order by which the resolvers are invoked. Here is how resolvers are registered: https://github.com/apache/tomcat/blob/c2ad60927672c0c2080bdfbf8274b152fa630312/java/jakarta/el/StandardELContext.java#L53-L58

Also the way to register custom types is via registering ELResolver to the ELContext (notice that [TypeConverter](https://github.com/apache/tomcat/blob/d27c63055a679fb461f88ed07acd9b6962fac187/java/jakarta/el/TypeConverter.java#L22) extends ELResolver resolver and type converter implement . If I simply return EL support (from the code path of compute step), this code will end up in infinite loop:
https://github.com/apache/tomcat/blob/d27c63055a679fb461f88ed07acd9b6962fac187/java/jakarta/el/CompositeELResolver.java#L150

aymkhalil: if return null is confusing, I can separate the code paths for EL context aware (like compute step) and context non-aware (for cast, or direct invocation of the converter from utility methods) paths.

aymkhalil:
In the latest patch:
* I disabled method invocations by implementing a custom EL context 
* Separated the code paths for `coerceTo*` methods on the `CustomTypeConverter`
@cbornet PTAL

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/CustomTypeConverter.java
aymkhalil: I used a variant of option 1 in the latest path where the Jstl conversions return null and direct converter invocation (e.g. from cast step or utility methods) overrides the Jstl class and handles the null by returning `ELSupport.coerce*` methods

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/DisabledInvocationBeanResolver.java
```diff
@@ -0,0 +1,28 @@
...
+
+import jakarta.el.BeanELResolver;
+import jakarta.el.ELContext;
+import jakarta.el.MethodNotFoundException;
+
+public class DisabledInvocationBeanResolver extends BeanELResolver {
+  @Override
+  public Object invoke(
+      ELContext context, Object base, Object method, Class<?>[] paramTypes, Object[] params) {
+    throw new MethodNotFoundException("Method invocations are disabled: " + method);
```
eolivelli, line 26: great idea !


Maybe it would be useful to add a comment to explain why we are doing so

eolivelli (approved):

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/StandardContext.java
```diff
@@ -0,0 +1,176 @@
...
+ * A standard context that mirrors {@link jakarta.el.StandardELContext} with the exception that it
+ * registers a custom beans resolver that disables invocations.
+ */
+public class StandardContext extends ELContext {
+  private final ELContext wrappedContext;
+  private final VariableMapper variableMapper;
+  private final FunctionMapper functionMapper;
+  private final CompositeELResolver standardResolver;
+  private final CompositeELResolver customResolvers;
+  private final Map<String, Object> localBeans = new HashMap();
```
cbornet, line 47:
```suggestion
  private final Map<String, Object> localBeans = new HashMap<>();
```

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/StandardContext.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  public VariableMapper getVariableMapper() {
+    return this.variableMapper;
+  }
+
+  Map<String, Object> getLocalBeans() {
+    return this.localBeans;
+  }
+
```
cbornet, line 104:
```suggestion
```

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/StandardContext.java
```diff
@@ -0,0 +1,176 @@
...
+  private static class StandardFunctionMapper extends FunctionMapper {
+    private final Map<String, Method> methods = new HashMap();
```
cbornet, line 106:
```suggestion
    private final Map<String, Method> methods = new HashMap<>();
```

cbornet:
Since the custom resolvers are used before the default ones, I think we can just add `DisabledInvocationBeanResolver` to `StandardELContext`. And thus remove `StandardContext`.
I tried this code in `JstlEvaluator`
```java
    final StandardELContext standardContext = new StandardELContext(FACTORY);
    standardContext.addELResolver(new DisabledInvocationBeanResolver());
    standardContext.addELResolver(JstlTypeConverter.INSTANCE);
```
and the tests pass.
WDYT ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/StandardContext.java
```diff
@@ -0,0 +1,176 @@
...
+
+    public StandardFunctionMapper(Map<String, Method> initFunctionMap) {
+      if (initFunctionMap != null) {
+        this.methods.putAll(initFunctionMap);
+      }
+    }
+
+    public Method resolveFunction(String prefix, String localName) {
+      String key = prefix + ":" + localName;
+      return (Method) this.methods.get(key);
```
cbornet, line 116:
```suggestion
      return this.methods.get(key);
```

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/StandardContext.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+  }
+
+  private static class StandardVariableMapper extends VariableMapper {
+    private Map<String, ValueExpression> vars;
+
+    private StandardVariableMapper() {}
+
+    public ValueExpression resolveVariable(String variable) {
+      return this.vars == null ? null : (ValueExpression) this.vars.get(variable);
```
cbornet, line 163:
```suggestion
      return this.vars == null ? null : this.vars.get(variable);
```

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/StandardContext.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    public ValueExpression setVariable(String variable, ValueExpression expression) {
+      if (this.vars == null) {
+        this.vars = new HashMap();
+      }
+
+      return expression == null
+          ? (ValueExpression) this.vars.remove(variable)
+          : (ValueExpression) this.vars.put(variable, expression);
```
cbornet, line 173:
```suggestion
          ? this.vars.remove(variable)
          : this.vars.put(variable, expression);
```

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/StandardContext.java
```diff
@@ -0,0 +1,176 @@
...
+
+    private StandardVariableMapper() {}
+
+    public ValueExpression resolveVariable(String variable) {
+      return this.vars == null ? null : (ValueExpression) this.vars.get(variable);
+    }
+
+    public ValueExpression setVariable(String variable, ValueExpression expression) {
+      if (this.vars == null) {
+        this.vars = new HashMap();
```
cbornet, line 168:
```suggestion
        this.vars = new HashMap<>();
```

cbornet (approved):

PR #47, opened 2023-01-14
Use custom converter in dateadd function

cbornet:


tests/src/test/java/com/datastax/oss/pulsar/functions/transforms/tests/AbstractDockerTest.java
```diff
@@ -105,6 +107,70 @@ public void testPrimitive() throws Exception {
     assertEquals(value.getNativeObject(), "42");
   }
 
+  @Test
+  public void test1() throws Exception {
```
aymkhalil, line 111: CI is failing on this.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
```diff
@@ -67,33 +66,13 @@ public static long now() {
...
-    Instant instant = Instant.ofEpochMilli(epochMillis);
-    OffsetDateTime localDateTime = OffsetDateTime.ofInstant(instant, ZoneOffset.UTC);
-    return dateadd(localDateTime, delta, unit);
-  }
-
-  private static long dateadd(OffsetDateTime offsetDateTime, long delta, String unit) {
-    final ChronoUnit chronoUnit;
-    switch (unit) {
+    ChronoUnit chronoUnit;
+    switch (CustomTypeConverter.INSTANCE.coerceToString(unit)) {
```
aymkhalil, line 75: this way, `unit` is more permissive than expected for this use case. what's wrong with limiting it `String` type in the method signature and just using it as is?

aymkhalil (changes requested):
Mainly CI is failing, also was not sure what use case required our custom string coersion for the `unit` part

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
cbornet: It's not really more permissive, the values allowed are still the same. It just allows to use a type that is compatible.
Practically, only BYTES can match.

tests/src/test/java/com/datastax/oss/pulsar/functions/transforms/tests/AbstractDockerTest.java
cbornet: Oups. This should not have been committed.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
```diff
@@ -119,13 +98,25 @@ private static long dateadd(OffsetDateTime offsetDateTime, long delta, String un
       case "millis":
         chronoUnit = ChronoUnit.MILLIS;
         break;
+      case "nanos":
```
cbornet, line 101: Added nanos to the party.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlFunctions.java
```diff
@@ -119,13 +98,25 @@ private static long dateadd(OffsetDateTime offsetDateTime, long delta, String un
...
+    if (chronoUnit == ChronoUnit.MONTHS || chronoUnit == ChronoUnit.YEARS) {
+      return CustomTypeConverter.INSTANCE
+          .coerceToOffsetDateTime(input)
+          .plus(CustomTypeConverter.INSTANCE.coerceToLong(delta), chronoUnit)
+          .toInstant()
+          .toEpochMilli();
+    }
+    return CustomTypeConverter.INSTANCE
+        .coerceToInstant(input)
+        .plus(CustomTypeConverter.INSTANCE.coerceToLong(delta), chronoUnit)
```
cbornet, line 119: `plus` operation is faster on Instant.

cbornet:
The unit conversion is because the string could come from a BYTES SchemaType.

PR #50, opened 2023-01-17
Infer type from expression if it's not explicitly set

cbornet:
With this change it's not mandatory to provide the "type" field in the compute step. It will be inferred from the result of the evaluation.
Note that the EL spec is strict on the return type of the expressions (eg: arithmetic operations always return long or double, even if they are done on int or float). So there's no risk of a change of return type between lib version or lib change.


cbornet:
How do we handle expressions evaluating to null ?

aymkhalil:
> How do we handle expressions evaluating to null ?

I think with inferred types, it is fair to throw an exception when null. 
maybe we should encourage the user to use:
```fn:coalesce(input, 'default')``` so we always have a value to infer from, or just use explicit types!

File: pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/ComputeStepTest.java [hunk 1]
```diff
@@ -225,6 +222,106 @@ void testAvro() throws Exception {
     assertEquals(read.get("age"), new Utf8("43"));
   }
 
+  @Test
+  void testAvroInferredType() throws Exception {
```
aymkhalil, line 226: However we handle expression that evaluates to null, it has to be covered with a unit test (I think it is fair to throw an exception as mentioned in the comment).

aymkhalil (approved):

PR #55, opened 2023-01-18
Coerce Date to epoch days

aymkhalil:
Convert `j.u.Date` to epoch days when applicable. This way it is consistent with AVRO's date logical type.

### DIFF
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlTypeConverter.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlTypeConverter.java
@@ -31,6 +31,7 @@ import java.time.format.DateTimeFormatter;
 import java.time.temporal.TemporalAccessor;
 import java.util.Date;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 import org.apache.avro.util.Utf8;
 import org.apache.el.util.MessageFactory;
 import org.apache.pulsar.client.api.Schema;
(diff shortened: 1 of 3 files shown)


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlTypeConverter.java
```diff
@@ -107,6 +115,12 @@ protected Integer coerceToInteger(Object value) {
     if (value instanceof Time) {
       return (int) (((Time) value).toLocalTime().toNanoOfDay() / 1_000_000);
     }
+    if (value instanceof Date) {
```
aymkhalil, line 118: Or should I use `Date.class.equals(value.getClass()))` to avoid matching a Timestamp?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/jstl/JstlTypeConverter.java
```diff
@@ -88,8 +90,14 @@ protected Long coerceToLong(Object value) {
     if (value instanceof Time) {
       return ((Time) value).toLocalTime().toNanoOfDay() / 1_000_000;
     }
+    if (value instanceof Timestamp) {
+      return ((Timestamp) value).getTime();
+    }
     if (value instanceof Date) {
-      return ((Date) value).getTime();
+      return ((Date) value).getTime() / MILLIS_PER_DAY;
```
cbornet, line 97: I think j.u.Date should stay as a timestamp. We only do the change for LocalDate

PR #56, opened 2023-01-18
Use number of days since epoch in LocalDate conversions

cbornet:


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeStep.java
```diff
@@ -293,8 +293,8 @@ private Long getAvroTimestampMillis(Object value, LogicalType logicalType) {
   private Integer getAvroDate(Object value, LogicalType logicalType) {
     validateLogicalType(value, logicalType, Date.class, LocalDate.class);
     return (value instanceof LocalDate)
-        ? (int) ((LocalDate) value).toEpochDay()
-        : (int) (((Date) value).getTime() / MILLIS_PER_DAY);
+        ? Math.toIntExact(((LocalDate) value).toEpochDay())
```
aymkhalil, line 296: Max year that can be represented by epoch days as integer is "5881580-07-11"! After that, our transformation we overflow :D

aymkhalil (approved):

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeStep.java
cbornet: Yes... I prefer to fail here than outputing a random value.

aymkhalil: I'd be concerned about the world if this code is still in use by the year 5881580

PR #58, opened 2023-01-18
Update the coerce methods error codes to match tomcat-el-embeded

aymkhalil:
Matching the error codes used in ELSupport. E.g. https://github.com/apache/tomcat/blob/e950ca41e03353a753960230b488e9723c6f5782/java/org/apache/el/lang/ELSupport.java#L334


aymkhalil:
The full list of localized strings is defined here: https://github.com/apache/tomcat/blob/1da1552abc56a46e6c46113f42fa2624894e2dc5/java/org/apache/el/LocalStrings.properties

PR #59, opened 2023-01-18
Change DATE to map to number of days since epoch

cbornet:


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
```diff
@@ -319,6 +319,13 @@ private static TransformStep newComputeFieldFunction(ComputeConfig config) {
                 throw new IllegalArgumentException(
                     "Duplicate compute field name detected: " + field.getName());
               }
+              if (field.getType() == ComputeFieldType.DATE
```
aymkhalil, line 322: Is there a need to update the type inference logic to explicitly not infer `DATE` for primitive types?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/ComputeField.java
```diff
@@ -142,7 +141,6 @@ private Class<?> getJavaType() {
         case BOOLEAN:
           return Boolean.class;
         case DATE:
-          return Date.class;
```
aymkhalil, line 145: Good thing is primitive types support on compute is not released yet so no one could be using `DATE` + primitive types so we should be safe here.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
```diff
@@ -319,6 +319,13 @@ private static TransformStep newComputeFieldFunction(ComputeConfig config) {
...
+                  && ("value".equals(field.getName()) || "key".equals(field.getName()))) {
+                throw new IllegalArgumentException(
```
aymkhalil, line 324: Would the user get this feedback when they create the function via pulsar admin - or just later on in the logs when the function actually initializes?

aymkhalil (approved):

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
cbornet: We don't have any function that would infer a DATE atm. And I believe there won't ever be.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
cbornet: In the logs.

Issue #62, opened 2023-02-15
[compute] Cannot convert Avro Utf8 string to INT32


nicoloboschi:
I have the input schema which is key-value with avro's value with a text field (`myvalue`). I want to convert it to INT (I assume that it's a valid number). This is my config for the function:
```
{"steps":[{"type":"compute","fields": 
       {"name":"value.value","expression":"value.myvalue","type":"INT32"}
]}
```

I'm getting this error
```
jakarta.el.ELException: Cannot convert [13360] of type [class org.apache.avro.util.Utf8] to [class java.lang.Integer]
	at org.apache.el.lang.ELSupport.coerceToNumber(ELSupport.java:390) ~[?:?]
	at org.apache.el.lang.ELSupport.coerceToType(ELSupport.java:532) ~[?:?]
	at org.apache.el.ExpressionFactoryImpl.coerceToType(ExpressionFactoryImpl.java:46) ~[?:?]
	at jakarta.el.ELContext.convertToType(ELContext.java:320) ~[?:?]
	at org.apache.el.ValueExpressionImpl.getValue(ValueExpressionImpl.java:192) ~[?:?]
	at com.datastax.oss.pulsar.functions.transforms.jstl.JstlEvaluator.evaluate(JstlEvaluator.java:119) ~[?:?]
	at com.datastax.oss.pulsar.functions.transforms.ComputeStep.computeFields(ComputeStep.java:250) ~[?:?]
	at com.datastax.oss.pulsar.functions.transforms.ComputeStep.computeValueFields(ComputeStep.java:86) ~[?:?]
	at com.datastax.oss.pulsar.functions.transforms.ComputeStep.process(ComputeStep.java:69) ~[?:?]
	at com.datastax.oss.pulsar.functions.transforms.TransformFunction.process(TransformFunction.java:261) ~[?:?]
	at com.datastax.oss.pulsar.functions.transforms.TransformFunction.process(TransformFunction.java:251) ~[?:?]
	at com.datastax.oss.pulsar.functions.transforms.TransformFunction.process(TransformFunction.java:123) ~[?:?]
	at org.apache.pulsar.functions.instance.JavaInstance.handleMessage(JavaInstance.java:94) ~[com.datastax.oss-pulsar-functions-instance-2.10.3.2.jar:2.10.3.2]
[...]

PR #63, opened 2023-02-16
Fix Utf8 coercions

aymkhalil:
**Fixes** #62 

**Implementation notes**
In order to take advantage of our customer Utf8 coercions, this patch directly calls ELSupport functions instead of relying on the ExpressionFactoryImp to do the coercion. Although it is possible to extend ExpressionFactoryImp and make it Utf8 aware, the tomcat-embeded-el runtime instantiates its own ExpressionFactory via [static](https://github.com/apache/tomcat/blob/main/java/jakarta/el/ExpressionFactory.java#L62) method which makes it hard to inject our custom expression factory. However, we have full control over the custom converter injected in StandardContext.

This patch also removes `CustomTypeConverter` in favor of `JstlTypeConverter` which turned out to be simpler 

**Testing**
Added `testComputeUtf8ToInteger` integ test


eolivelli (approved):
Makes sense to me.

The test shows that we are fixing the problem

PR #68, opened 2023-04-04
Fix resolving AVRO logical types in JstlTransformContextAdapter

### LINKED ISSUE #66: [compute][logical types] Coercion of logical types is not working as expected.
In a C* CDC -> Sink context, I was trying to convert an Avro date (which is a logical type) to string, and the output was a string representing the number of days since epoch. The expected output should be a an iso formated date 'DDDD-YY-MM'.

**Example input:** Use this topic schema to reproduce 
```
{
  "key": {
    "type": "record",
    "name": "test_date",
    "namespace": "ks1",
    "doc": "Table ks1.test_date",
    "fields": [
      {
        "name": "id",
        "type": "string"
      }
    ]
  },
  "value": {
    "type": "record",
    "name": "test_date",
    "namespace": "ks1",
    "doc": "Table ks1.test_date",
    "fields": [
      {
        "name": "val",
        "type": [
          "null",
          {
            "type": "int",
            "logicalType": "date"
          }
        ],
        "default": null
      }
    ]
  }
}
```

And then the sink was created with the following transform function:
```
{
   "steps":[
      {
         "type":"compute",
         "fields":[
            {
               "name":"value.val",
               "expression":"value.val",
               "type":"STRING"
            }
         ]
      }
   ]
}
```

and the output of the transformation (shown in the sink error) was:
```
com.datastax.oss.sink.pulsar.CassandraSinkTask - Error decoding/mapping Pulsar record PulsarSinkRecord{OutputRecordSinkRecord(sinkRecord=org.apache.pulsar.functions.api.utils.FunctionRecord@6a7cb571, value=(key = "org.apache.pulsar.client.impl.schema.generic.GenericAvroRecord@27ca9596", value = "org.apache.pulsar.client.impl.schema.generic.GenericAvroRecord@5c6acc98"), schema=KeyValueSchema(SEPARATED,org.apache.pulsar.client.impl.schema.generic.GenericAvroSchema@524e0232,org.apache.pulsar.client.impl.schema.generic.GenericAvroSchema@3091f9fe))}: Text '15402' could not be parsed at index 0
```

The relevant part is "**Text '15402' could not be parsed at index 0**". 

The expected text value should've been `2012-03-03`

aymkhalil:
Fixes #66 

This patch fixes how the JstlTransformContextAdapter fetches AVRO logical type info. Before, it was working only if the logical type was inside a non-union schema. In general, Avro nullable types (like those produced by the compute step) would have a union schema like the following (for those, logical types won't be found by simply accessing `schema().getLogicalType()` and the nested types should be consulted):
```
{
        "name": "...",
        "type": [
          "null",
          {
            "type": "...",
            "logicalType": "..."
          }
        ],
        "default": null
      }
```

### DIFF
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeStep.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeStep.java
@@ -20,6 +20,7 @@ import static org.apache.pulsar.common.schema.SchemaType.AVRO;
 import com.datastax.oss.pulsar.functions.transforms.jstl.JstlTypeConverter;
 import com.datastax.oss.pulsar.functions.transforms.model.ComputeField;
 import com.datastax.oss.pulsar.functions.transforms.model.ComputeFieldType;
+import com.datastax.oss.pulsar.functions.transforms.util.AvroUtil;
 import java.nio.ByteBuffer;
 import java.sql.Time;
 import java.sql.Timestamp;
@@ -32,7 +33,6 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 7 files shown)


eolivelli (approved):
+1

awesome work

PR #70, opened 2023-06-19
Add step for generate AI embeddings

nicoloboschi:
- Added new "compute-ai-embeddings" step
- It relies on the openai config implemented in https://github.com/datastax/pulsar-transformations/pull/71
- It adds/fill a new field `embedding-field` into the record value with the embeddings of the specified `fields`. The `model` must specifed.

Config example:
```
--user-config '{"openai": {"url": "https://datastax-openai-dev.openai.azure.com", "access-key": "KEY", "provider": "azure"}, "steps": [{"type": "compute-ai-embeddings", "model": "text-embedding-ada-002", "fields": ["field1"], "embeddings-field": "embeddings"}]}'
```

Supported providers:
- OpenAI
- Azure

Record schema supported:
- [x] AVRO
- [x] JSON
- [x] Key Value (With structured value) 

Notes:
- Each model has a max input size (calculated after the [tokenization](https://platform.openai.com/docs/api-reference/embeddings/create#embeddings/create-input)). If the input exceeds the limit, the function will fail. We could support in the future the [chunking of the inputs.](https://github.com/openai/openai-cookbook/blob/main/examples/Embedding_long_inputs.ipynb)
- In case of multiple input fields specified, the field values are concatenated and sent as single token. Dot product is not supported.

Open design questions:
- Should we allow templating the output value as suggested by @cbornet https://github.com/datastax/pulsar-transformations/pull/70#issuecomment-1598539793 ? 
- What do we do with primitive schema ? Currently they are skipped. Also, the `cast` step doesn't support moving from a primitive schema to a structured one (json,avro).


tests/pom.xml
```diff
@@ -65,7 +65,7 @@
             <configuration>
               <tasks>
                 <echo>copy proxy protocol handler</echo>
-                <copy file="${basedir}/../pulsar-transformations/target/pulsar-transformations-${project.version}.nar" tofile="${project.build.outputDirectory}/pulsar-transformations.nar" />
```
cbornet, line 68: We should avoid committing this one.
I don't know why it goes always adding/removing the space.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
```diff
@@ -225,6 +228,9 @@ public void initialize(Context context) {
         case "compute":
           transformStep = newComputeFieldFunction((ComputeConfig) step);
           break;
+        case "add-embeddings":
```
eolivelli, line 231: what about `compute-ai-embeddings` ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/EmbeddingsClient.java
```diff
@@ -0,0 +1,8 @@
+package com.datastax.oss.pulsar.functions.transforms.embeddings;
+
+import java.util.List;
+
+public interface EmbeddingsClient {
+
+  List<List<Double>> calculateEmbeddings(List<String> texts);
```
eolivelli, line 7: "throws" ?

eolivelli (commented):
Overall it seems a good starting point (of course we have to handle multiple fields....)

One higher level question:
how can we make this work with a JSON string (without a Schema in the topic) ?
do we have a step that converts from a raw string to a Struct ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
aymkhalil: actually if the is a compute kind of transformation how about leveraging the compute step and add more JSTL functions under a new `ai:` prefix?

For example, `ai:embed(input)` to invoke vector embeddings or `ai:complete()` to invoke the completion API. With that approach, you get all the compute features for free and the UX is consistent. We'd just need away to pass the authorization and the api endpoint to invoke.

Here is how the transformation could look like:

```
{
    "steps": [
        {
            "type": "compute",
            "fields": [
                {
                    "name": "value.vectorizedText",
                    "type": "STRING",
                    "expression": "ai:embed(value.text)"
                }
            ]
        }
    ]
}
```

eolivelli: I think that it is better to have a new step, because it will make the feature more visible to users

pom.xml.releaseBackup
```diff
@@ -0,0 +1,369 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
eolivelli, line 1: this file shouldn't be committed

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/EmbeddingsService.java
```diff
@@ -0,0 +1,8 @@
+package com.datastax.oss.pulsar.functions.transforms.embeddings;
+
+import java.util.List;
+
+public interface EmbeddingsService {
+
+  List<List<Double>> calculateEmbeddings(List<String> texts);
```
eolivelli, line 7: "throws" ?

Also ideally this is a async API, currently Transforms aren't supporting async processing so we can keep it this way, but in the future we have to make it fully async, otherwise it is hard to deal with parallelism

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/OpenAIEmbeddingsService.java
```diff
@@ -0,0 +1,30 @@
...
+  private final String model;
+
+  public OpenAIEmbeddingsService(OpenAIClient openAIClient, String model) {
+    this.openAIClient = openAIClient;
+    this.model = model;
+  }
+
+  @Override
+  public List<List<Double>> calculateEmbeddings(List<String> texts) {
+    EmbeddingsOptions embeddingsOptions = new EmbeddingsOptions(texts);
```
eolivelli, line 22: we should add metrics one day here

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/EmbeddingsService.java
nicoloboschi: I think there s no need here for a checked exception for now. In the future we might want to let the user configures the behavior in case of too long inputs. In that case we could implement a specific exception

I'll keep it as sync for now

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/OpenAIEmbeddingsService.java
nicoloboschi: good point, we can do it in a follow up pull request. Currently there's no metric for any transformations. 
we could add a generic implementation per-step and let the step adds more specific metrics 

Unfortuantely the current Pulsar Functions Framework only supports counters. In this case it would be great to measure latency/thoughrput of the openAI API calls

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/OpenAIEmbeddingsService.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public OpenAIEmbeddingsService(OpenAIClient openAIClient, String model) {
+    this.openAIClient = openAIClient;
+    this.model = model;
+  }
+
+  @Override
+  public List<List<Double>> computeEmbeddings(List<String> texts) {
+    EmbeddingsOptions embeddingsOptions = new EmbeddingsOptions(texts);
+    Embeddings embeddings = openAIClient.getEmbeddings(model, embeddingsOptions);
```
cbornet, line 37: The client uses the term `deploymentId` instead of `model`.
Which should we use. At least we must be consistent between Embeddings and Completion

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/config/ComputeAIEmbeddingsConfig.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package com.datastax.oss.pulsar.functions.transforms.model.config;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import lombok.Getter;
+
+@Getter
+public class ComputeAIEmbeddingsConfig extends StepConfig {
+  @JsonProperty(value = "model", required = true)
```
cbornet, line 24: deploymentId ?

cbornet:
> It adds/fill a new field embedding-field into the record value with the embeddings of the specified fields. The model must specifed.

I think it would be nice to be able to apply a template to the record. Then the user can have more complex embedding inputs.
That's what I'm doing for the chat completions.
Something like
```json
"embeddings": [
    "The name of the user is {{ value.user }}",
    "The destination topic is {{ destinationTopic }}"
]
```

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeAIEmbeddingsStep.java
```diff
@@ -0,0 +1,174 @@
...
+      if (field.name().equals(embeddingsFieldName)) {
+        newRecord.put(field.name(), embeddings);
+      } else {
+        newRecord.put(field.name(), avroRecord.get(field.name()));
+      }
+    }
+    if (avroRecord != newRecord) {
+      transformContext.setValueModified(true);
+    }
+    transformContext.setValueObject(newRecord);
```
cbornet, line 120: Don't you need to configure the schema ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeAIEmbeddingsStep.java
```diff
@@ -0,0 +1,174 @@
...
+            .stream()
+            .map(f -> values.get(f))
+            .filter(Objects::nonNull)
+            .collect(Collectors.joining(" "));
+
+    final List<Double> embeddings = embeddingsService.computeEmbeddings(List.of(text)).get(0);
+    applyEmbeddingsToRecord(transformContext, embeddings);
+  }
+
+  private void applyEmbeddingsToRecord(TransformContext transformContext, List<Double> embeddings) {
```
cbornet, line 70: I think the first release could be to output the value with Schema.STRING

cbornet:
Actually, we only compute 1 embedding so that would be
```json
"text": "The name of the user is {{ value.user }}, the destination topic is {{ destinationTopic }} and the message key is {{ key }}"
```

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/OpenAIEmbeddingsService.java
nicoloboschi: In the API is called "model". https://platform.openai.com/docs/api-reference/embeddings/create#embeddings/create-model
It is referred as model in the OpenAI doc: https://platform.openai.com/docs/models/embeddings

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeAIEmbeddingsStep.java
nicoloboschi: it's a compute transformation so IMHO the expectation is to have the same message enriched with the embeddings. 
Actually, in case of primitive schemas, we could output the embeddings instead of the value.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeAIEmbeddingsStep.java
nicoloboschi: It's handled in the `TransformContext#send` method. Do you think we need additional logic ?

nicoloboschi:
> > It adds/fill a new field embedding-field into the record value with the embeddings of the specified fields. The model must specifed.
> 
> I think it would be nice to be able to apply a template to the record. Then the user can have more complex embedding inputs. That's what I'm doing for the chat completions. Something like
> 
> ```json
> "texts": [
>     "The name of the user is {{ value.user }}",
>     "The destination topic is {{ destinationTopic }} and the message key is {{ key }}"
> ]
> ```

It makes sense. We could add it later

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/config/ComputeAIEmbeddingsConfig.java
nicoloboschi: see above

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/OpenAIEmbeddingsService.java
cbornet: yes but why Azure called it deploymentId then ?

eolivelli:
> > It makes sense. We could add it later
> 
> I think it should be done from the start. Otherwise there will be backward compatibility issues.

We are in a prototype mode
it's fine to have it this way

PR #74, opened 2023-06-20
Add chat completions step

cbornet:


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/config/StepConfig.java
```diff
@@ -30,6 +30,7 @@
     @JsonSubTypes.Type(value = DropConfig.class, name = "drop"),
     @JsonSubTypes.Type(value = FlattenConfig.class, name = "flatten"),
     @JsonSubTypes.Type(value = ComputeConfig.class, name = "compute"),
+    @JsonSubTypes.Type(value = ChatCompletionsConfig.class, name = "chat-completions"),
```
eolivelli, line 33: what about using a name beginning with "ai-" ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/config/ChatCompletionsConfig.java
```diff
@@ -0,0 +1,56 @@
...
+package com.datastax.oss.pulsar.functions.transforms.model.config;
+
+import com.azure.ai.openai.models.ChatMessage;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import java.util.Map;
+import lombok.Data;
+
+@Data
+public class ChatCompletionsConfig extends StepConfig {
```
aymkhalil, line 25: Can the user select a certain field on the topic as input to the completion step?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ChatCompletionsStep.java
```diff
@@ -0,0 +1,120 @@
...
+        return val;
+      default:
+        throw new UnsupportedOperationException(
+            "Unsupported schemaType " + schema.getSchemaInfo().getType());
+    }
+  }
+
+  private String applyTemplate(String template, JsonRecord record) {
+    // TODO: pre-compile the templates at init time
+    Template tmpl = Mustache.compiler().compile(template);
```
aymkhalil, line 117: Is it possible to extend the Mustache template to leverage `JstlTransformContextAdapter`? This would enable users to use the `fn:` functions or evaluate any field the way compute does it. If not, at least we need to make sure that the `dot` notation to access fields is consistent with compute.

eolivelli: this is a great idea

eolivelli (approved):

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/config/ChatCompletionsConfig.java
cbornet: Yes. Mustache templating can be used in the chat messages.
The model used for templating is
```java
@Data
public class JsonRecord {
  String topicName;
  String destinationTopic;
  Object key;
  Object value;
  Map<String, String> properties;
  Long eventTime;
}
``` 
so you can do `The topic is {{ destinationTopic }}.`
If value/key is AVRO in the record, they are converted to `Map<String, Object>` in the JsonRecord.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ChatCompletionsStep.java
cbornet: Mustache uses the `dot` annotation too.
I don't see how it can be mixed with JstlTransformContextAdapter

aymkhalil: I was thinking if Mustache allows for custom evaluations so first level is to detect anything between `{{}}` and the second step is to delegate to `JstlTransformContextAdapter` - on the second thought this may confuse the user because the syntax will be mixed between the two - maybe better to stick with Mustache

PR #76, opened 2023-06-21
[query] Add new Query Step to perform queries to external databases (with AstraDB Vector Search)

eolivelli:
- new "query" step
- initial implementation for AstraDB (using the DataStax cassandra driver)

Sample configuration:

```
{
  "datasource": {
    "service": "astra",
    "username": "........",
    "password": "....",
    "secureBundle": "UEsDBBQACAAIANRT1VYAAAAA....."
  },
  "steps": [
    {
      "type": "query",
      "query": "SELECT * FROM vsearch.products where id=?",
      "output-field": "results",
      "fields": ["id"]
    }
  ]
}
```

Sample configuration to compute a similarity search on a Vector Database (AstraDB)

```
{
  "datasource": {
    "service": "astra",
    "username": ".....",
    "password": "......t",
    "secureBundle": "......"
  },
  "openai": {"url": "https://datastax-openai-dev.openai.azure.com/", "access-key": "......", "provider": "azure"},
  "steps": [
    {
      "type": "compute-ai-embeddings",
      "model": "text-embedding-ada-002",
      "embeddings-field": "embeddings",
      "fields": ["name", "description"]
    },
    {
      "type": "query",
      "query": "SELECT * FROM vsearch.products ORDER BY item_vector ANN OF ? LIMIT 1;",
      "output-field": "results",
      "fields": ["embeddings"]
    }
  ]
}
```


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/QueryStep.java
```diff
@@ -0,0 +1,244 @@
...
+/**
+ * Compute AI Embeddings for one or more records fields and put the value into a new or existing
+ * field.
+ */
+@Builder
+@Slf4j
+public class QueryStep implements TransformStep {
+
+  @Builder.Default private final List<String> fields = new ArrayList<>();
+  @Builder.Default private final String outputFieldName;
```
cbornet, line 46: I think you need to provide a default value if you annotate with `Builder.Default`

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/QueryStep.java
```diff
@@ -0,0 +1,244 @@
...
+
+  private void applyResultsToRecord(
+      TransformContext transformContext, List<Map<String, String>> results) {
+    final SchemaType type = transformContext.getValueSchema().getSchemaInfo().getType();
+    switch (type) {
+      case AVRO:
+        applyResultsToAvroRecord(transformContext, results);
+        break;
+      case JSON:
+        applyResultsToJsonRecord(transformContext, results);
```
cbornet, line 73: We don't really support JSON in transformations today.
This would need to handle the schema.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/QueryStep.java
```diff
@@ -0,0 +1,244 @@
...
+        break;
+      default:
+        return;
+    }
+  }
+
+  private void applyResultsToAvroRecord(
```
cbornet, line 80: Should use `TransformContext::addOrReplaceAvroValueFields`

pulsar-transformations/src/main/resources/config-schema.yaml
```diff
@@ -44,6 +44,7 @@ components:
           compute: "#/components/schemas/Compute"
           compute-ai-embeddings: "#/components/schemas/ComputeAIEmbeddings"
           ai-chat-completions: "#/components/schemas/AIChatCompletions"
+          query: "#/components/schemas/Query"
```
cbornet, line 47: need a name more explicit about what type of query it is

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/QueryStep.java
```diff
@@ -0,0 +1,244 @@
...
+            avroSchema,
+            schema -> {
+              AtomicBoolean fieldExists = new AtomicBoolean(false);
+              final List<Schema.Field> newFields =
+                  avroSchema
+                      .getFields()
+                      .stream()
+                      .peek(
+                          f -> {
+                            if (f.name().equals(outputFieldName)) {
```
cbornet, line 95: I think the logic in ChatCompletionsStep is more powerful than the one from EmbeddingsStep as it allows to put the results in a lot more places such as properties, key, full key or value, and so on.
So I would reuse the logic from there and also refactor EmbeddingsStep to do the same so we have consistency.

cbornet (changes requested):

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/QueryStep.java
```diff
@@ -0,0 +1,244 @@
...
+  private final Map<Schema, Schema> avroValueSchemaCache = new ConcurrentHashMap<>();
+
+  @Override
+  public void process(TransformContext transformContext) {
+    if (!isSchemaCompatible(transformContext)) {
+      return;
+    }
+
+    Map<String, Object> values = new HashMap<>();
+    collectFieldValuesFromKey(fields, transformContext, values);
```
cbornet, line 58: I think we should use a template like in ChatCompletions.
We should also do it for Embeddings.

cbornet: Well, actually not exactly a template is needed here but we should be able to use `value.field` or `destinationTopic` or `properties.foo` like we have in `compute`

pulsar-transformations/src/main/resources/config-schema.yaml
eolivelli: This is a "query", maybe we can name it "query-datasource" ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/QueryStep.java
eolivelli: I agree, we should refactor and use common methods, I would do it in a follow up step

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/QueryStep.java
eolivelli: I agree, we should refactor and use common methods, I would do it in a follow up step

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/QueryStep.java
eolivelli: I agree, we should refactor and use common methods, I would do it in a follow up step

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/QueryStep.java
eolivelli: I know, in fact it doesn't work.
In any case we should have come common methods

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/QueryStep.java
eolivelli: I think that the way we are using @Default is not good and I see many warning.
I have copied from other steps.

PR #78, opened 2023-06-22
Use a template filled with the message fields and metadata for the embeddings

cbornet:
You can have the same as before by doing
```json
{"steps": [{"type": "compute-ai-embeddings", "text": "{{ value.valueField1 }} {{ value.valueField2 }}", "embeddings-field": "emb", "model": "gpt-3.5-turbo"}]}
```


eolivelli (approved):
+1

PR #79, opened 2023-06-23
Add field parameter resolution in QueryStep coherent with compute step

cbornet:


eolivelli (approved):
+1
great

pulsar-transformations/src/main/resources/config-schema.yaml
```diff
@@ -42,8 +42,8 @@ components:
           flatten: "#/components/schemas/Flatten"
           drop: "#/components/schemas/Drop"
           compute: "#/components/schemas/Compute"
-          compute-ai-embeddings: "#/components/schemas/ComputeAIEmbeddings"
-          ai-chat-completions: "#/components/schemas/AIChatCompletions"
+          compute-ai-embeddings: "#/components/schemas/ComputeAiEmbeddings"
```
nicoloboschi, line 45: ok so the master was broken!! this is error I was having yesterday!

PR #82, opened 2023-06-23
Use key instead of messageKey in EL

cbornet:
* messageKey is still possible for backward compatibility
* if the shema is KeyValue, key refers to the key part of the KV. Otherwise key refers to the message key in metadata.


eolivelli (approved):
+1
Great that you are keeping backward compatibility

aymkhalil:
> * if the shema is KeyValue, key refers to the key part of the KV. Otherwise key refers to the message key in metadata.

@cbornet How would the user access both the `messageKey` and `key` in an INLINE encoded key value schema after this patch?

For example (output from consume CLI):
```
key:[my optional key], properties:[], content:{key={value=my encoded key, type=class java.lang.String}, value={embeddings=NULL, id=0, text=Honda Accord Claims the Title of Best-Selling Car in the United States}}
```

PR #84, opened 2023-06-23
EmbeddingsService implementations for Huggingface

dlg99:
EmbeddingsService implementations for Huggingface

* using rest api
* using local (file://.. uri) model

The model requested there should be trained for "sentence similarity" task

### DIFF
--- a/pulsar-transformations/pom.xml
+++ b/pulsar-transformations/pom.xml
@@ -31,6 +31,38 @@
   </properties>
   <dependencies>
     <dependency>
+      <groupId>ai.djl</groupId>
+      <artifactId>api</artifactId>
+      <version>${djl.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>ai.djl.pytorch</groupId>
+      <artifactId>pytorch-engine</artifactId>
+      <version>${djl.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>ai.djl.pytorch</groupId>
+      <artifactId>pytorch-jni</artifactId>
+      <version>2.0.0-${djl.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>ai.djl.pytorch</groupId>
+      <artifactId>pytorch-native-cpu</artifactId>
+      <!--classifier>osx-aarch64</classifier-->
+      <classifier>osx-x86_64</classifier>
+      <version>2.0.0</version>
+    </dependency>
+    <dependency>
+      <groupId>ai.djl.pytorch</groupId>
+      <artifactId>pytorch-model-zoo</artifactId>
+      <version>${djl.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>ai.djl.huggingface</groupId>
+      <artifactId>tokenizers</artifactId>
+      <version>${djl.version}</version>
+    </dependency>
+    <dependency>
       <groupId>${pulsar.groupId}</groupId>
       <artifactId>pulsar-functions-api</artifactId>
       <scope>provided</scope>
--- a/pulsar-transformations/pom.xml
+++ b/pulsar-transformations/pom.xml
@@ -34,16 +34,19 @@
       <groupId>ai.djl</groupId>
       <artifactId>api</artifactId>
       <version>${djl.version}</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>ai.djl.pytorch</groupId>
       <artifactId>pytorch-engine</artifactId>
       <version>${djl.version}</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>ai.djl.pytorch</groupId>
       <artifactId>pytorch-jni</artifactId>
       <version>2.0.0-${djl.version}</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>ai.djl.pytorch</groupId>
@@ -51,16 +54,19 @@
       <!--classifier>osx-aarch64</classifier-->
       <classifier>osx-x86_64</classifier>
       <version>2.0.0</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>ai.djl.pytorch</groupId>
       <artifactId>pytorch-model-zoo</artifactId>
       <version>${djl.version}</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>ai.djl.huggingface</groupId>
       <artifactId>tokenizers</artifactId>
       <version>${djl.version}</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>${pulsar.groupId}</groupId>
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
@@ -38,6 +38,7 @@ import com.datastax.oss.pulsar.functions.transforms.model.config.ComputeConfig;
 import com.datastax.oss.pulsar.functions.transforms.model.config.DataSourceConfig;
 import com.datastax.oss.pulsar.functions.transforms.model.config.DropFieldsConfig;
 import com.datastax.oss.pulsar.functions.transforms.model.config.FlattenConfig;
+import com.datastax.oss.pulsar.functions.transforms.model.config.HuggingFaceConfig;
 import com.datastax.oss.pulsar.functions.transforms.model.config.OpenAIConfig;
 import com.datastax.oss.pulsar.functions.transforms.model.config.OpenAIProvider;
 import com.datastax.oss.pulsar.functions.transforms.model.config.QueryConfig;
@@ -60,6 +61,7 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
@@ -149,6 +151,7 @@ public class TransformFunction
       Arrays.asList("value", "key", "destinationTopic", "messageKey", "topicName", "eventTime");
   private final List<StepPredicatePair> steps = new ArrayList<>();
   private OpenAIClient openAIClient;
+  private HuggingFaceConfig huggingConfig;
   private QueryStepDataSource dataSource;
 
   @Override
@@ -231,6 +234,7 @@ public class TransformFunction
     TransformStepConfig config = mapper.convertValue(userConfigMap, TransformStepConfig.class);
 
     openAIClient = buildOpenAIClient(config.getOpenai());
+    huggingConfig = config.getHuggingface();
     dataSource = buildDataSource(config.getDatasource());
 
     for (StepConfig step : config.getSteps()) {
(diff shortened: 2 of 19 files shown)


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
```diff
@@ -385,11 +391,38 @@ private static TransformStep newComputeFieldFunction(ComputeConfig config) {
...
+    ComputeAIEmbeddingsConfig.SupportedServices service =
+        ComputeAIEmbeddingsConfig.SupportedServices.valueOf(targetSvc.toUpperCase());
+
+    final EmbeddingsService embeddingService;
+    switch (service) {
+      case OPENAI:
+        embeddingService = new OpenAIEmbeddingsService(openAIClient, config.getModel());
+        break;
+      case HUGGINGFACES_REST:
+        String hfKey = System.getenv("huggingfaces.key");
```
eolivelli, line 410: Let's configure "huggingFace" like we do for "openai", with a top level configuration element
Because that's the way users can set their token.

The topic "secrets" should not be stored in configuration is a separate topic that in our services is tackled in a way that we cannot discuss here

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
```diff
@@ -385,11 +391,38 @@ private static TransformStep newComputeFieldFunction(ComputeConfig config) {
...
 
+  @SneakyThrows
   private TransformStep newComputeAIEmbeddings(ComputeAIEmbeddingsConfig config) {
+    String targetSvc = config.getService();
+    if (targetSvc == null || Strings.isNullOrEmpty(targetSvc)) {
+      targetSvc = ComputeAIEmbeddingsConfig.SupportedServices.OPENAI.name();
+    }
+
+    ComputeAIEmbeddingsConfig.SupportedServices service =
+        ComputeAIEmbeddingsConfig.SupportedServices.valueOf(targetSvc.toUpperCase());
```
eolivelli, line 402: let's add a level of smartness/autoconf here:
if the user configured only "openai" then the service here is "openai" by default,
if the user configured the HF API in the top level element, then the service is "hugging face api" by default

probably also for the "local" mode we can autodetect it somehow, it is not expected that a user uses both HG API and HG local mode in the same function.  (but it is expected that they use HG to compute embeddings and OpenAPI for Completion for instance)

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/HuggingFaceEmbeddingService.java
```diff
@@ -0,0 +1,74 @@
...
+      List<Double> l = new ArrayList<>(floats.length);
+      for (float aFloat : floats) {
+        l.add((double) aFloat);
+      }
+      out.add(l);
+    }
+    return out;
+  }
+
+  public static void main(String[] args) throws Exception {
```
eolivelli, line 60: please drop

maybe you can keep a disabled test in the "tests" folder, like I did for AstraDB

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/HuggingFaceRestEmbeddingService.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  public HuggingFaceRestEmbeddingService(HuggingRestConfig conf, String token, String model)
+      throws MalformedURLException {
+    this.conf = conf;
+    this.model = model;
+    this.token = token;
+    this.modelUrl = new URL(conf.hfUrl + model);
+
+    // TODO: try checking if model is valid https://huggingface.co/docs/datasets-server/valid
```
eolivelli, line 80: we should probably have initially ad hard coded whilelist of allowed models, this way we won't let users download anything without control.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/HuggingFaceRestEmbeddingService.java
```diff
@@ -0,0 +1,132 @@
...
+      Object result = om.readValue(body, Object.class);
+      return (List<List<Double>>) result;
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  // TODO: use some HTTP client library with better performance, if needed
+  private String query(String jsonPayload) throws Exception {
+    HttpURLConnection connection = (HttpURLConnection) modelUrl.openConnection();
```
eolivelli, line 102: IIRC the HttpURLConnection has a "shutdown" (or "close") method to be called in a finally block, in order to ensure that we don't leave dangling open connections in case of problems

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/HuggingFaceRestEmbeddingService.java
```diff
@@ -0,0 +1,132 @@
...
+    String line;
+    while ((line = reader.readLine()) != null) {
+      response.append(line);
+    }
+    reader.close();
+
+    return response.toString();
+  }
+
+  public static void main(String[] args) throws Exception {
```
eolivelli, line 124: please remove

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/config/ComputeAIEmbeddingsConfig.java
```diff
@@ -20,6 +20,13 @@
 
 @Getter
 public class ComputeAIEmbeddingsConfig extends StepConfig {
+
+  public enum SupportedServices {
+    OPENAI,
+    HUGGINGFACES_REST,
```
eolivelli, line 26: what about "HUGGINGFACE_API" ?

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/config/ComputeAIEmbeddingsConfig.java
```diff
@@ -20,6 +20,13 @@
...
+    HUGGINGFACES_DJL
```
eolivelli, line 27: what about HUGGINGFACE   ? (no "API", this is probably the mode that will be used most of the times, as it will provide better performances)

eolivelli (changes requested):
very good,
I have left some suggestions about the configuration and other nits

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
```diff
@@ -385,11 +391,38 @@ private static TransformStep newComputeFieldFunction(ComputeConfig config) {
     return ComputeStep.builder().fields(fieldList).build();
   }
 
+  @SneakyThrows
   private TransformStep newComputeAIEmbeddings(ComputeAIEmbeddingsConfig config) {
+    String targetSvc = config.getService();
+    if (targetSvc == null || Strings.isNullOrEmpty(targetSvc)) {
```
cbornet, line 397:
```suggestion
    if (Strings.isNullOrEmpty(targetSvc)) {
```

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/HuggingFaceRestEmbeddingService.java
```diff
@@ -0,0 +1,132 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.pulsar.functions.transforms.embeddings;
+
+import com.fasterxml.jackson.annotation.JsonAlias;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
```
cbornet, line 22: What is the minimum Java version we shall support ?
If it's 11 then we could use the new HttpClient

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/config/ComputeAIEmbeddingsConfig.java
```diff
@@ -28,4 +35,10 @@ public class ComputeAIEmbeddingsConfig extends StepConfig {
 
   @JsonProperty(value = "embeddings-field", required = true)
   private String embeddingsFieldName;
+
+  @JsonProperty(value = "compute-service")
+  private String service;
+
+  @JsonProperty(value = "compute-config-json")
+  private String configJson;
```
cbornet, line 43: This is going to be really painful to configure from the CLI (JSON in JSON with all the escapings...)
Can we use the HuggingConfig directly instead ?

cbornet (changes requested):

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/HuggingFaceRestEmbeddingService.java
eolivelli: JDK11 is the base.

if we use HttpClient than probably it makes sense to cache it in a field, it will allow better performances (and it also supports HTTP2)

+1

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/model/config/ComputeAIEmbeddingsConfig.java
eolivelli: most of the config (base URL for the API and the token) should be as a top level object, like "openai".

Here we keep only the other parameters specific to this step (like the "model" to use)

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformFunction.java
dlg99: OpenAI is configured either way because it is used in other steps, hence explicit use of config.getService().

pulsar-transformations/pom.xml
```diff
@@ -30,6 +30,38 @@
     <build.dir>${project.build.directory}</build.dir>
   </properties>
   <dependencies>
+    <dependency>
+      <groupId>ai.djl</groupId>
+      <artifactId>api</artifactId>
+      <version>${djl.version}</version>
```
eolivelli, line 36: please set this as "provided" and add the dependency in the new pulsar-ai-tools module.
Basically now we don't want the new third party dependencies to be bundled into the "pulsar-transformations" nar file but only in the ai-tools nar.

eolivelli (changes requested):
Can you please add some tests in TransformFunctionTest ?

for the HUGGINGFACE_API we can use wiremock
for the local mode execution I think that there are some models that are only a few KBs and we can commit them into the git repository (or use a git submodule)
https://huggingface.co/sentence-transformers
For instance this: https://huggingface.co/sentence-transformers/paraphrase-albert-small-v2


Regarding the configuration, I still think that from the user's point of view it is more elegant to have a configuration like the proposals below.
The main point is to make it easy to write the configuration.
I understand that from the code point of view this "flat" model may sound bad, but from the user's perspective it is easier to understand (and you have less intermediate JSON structs).

API mode:
```
{ 
    "huggingface": {
        "url": "https://...."     (useful for unit tests)
        "accessKey":"xxx",
        "mode": "api" 
    },
    "steps": [
       { type: "compute-ai-embeddings",
                    "model":"sentence-transformers/paraphrase-albert-small-v2"
                    "options":{}, // optional
                    "arguments":{}  // optional
       }
    ]
}
```

Local mode:

```
{ 
    "huggingface": {
        "mode": "compute-local" 
    },
    "steps": [
       { type: "compute-ai-embeddings",
                    "model":"sentence-transformers/paraphrase-albert-small-v2"
                    "options":{},  // optional
                    "arguments":{}  // optional 
       }
    ]
}
```

For "engine", if we test and support only PyTorch then we can drop the configuration parameter

eolivelli (commented):
Good,
I think that there is some json schema file to update in order to implement json validation of the configuration.

You can check the PRs that introduced the new steps

eolivelli (approved):
+1

PR #85, opened 2023-06-24
Add a module pulsar-ai-functions to have a distinct NAR for AI ops

cbornet:
The name of the built-in function will be `ai-functions`


pulsar-ai-functions/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-inline</artifactId>
+    </dependency>
+  </dependencies>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.nifi</groupId>
+        <artifactId>nifi-nar-maven-plugin</artifactId>
+        <version>1.2.0</version>
```
eolivelli, line 90: what about taking this chance to upgrade to 1.5.1 ?

pulsar-ai-functions/src/main/resources/META-INF/services/pulsar-io.yaml
```diff
@@ -0,0 +1,22 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+name: ai-functions
```
eolivelli, line 20: I don't have a better name at the moment, we can discuss with the team.
I think that we should have something like "gen-ai" or "ai-tools", ideally we should get rid of the "function" term

eolivelli (changes requested):
Very good

Only one comment before committing the patch:
 
It would be great to not bundle the OpenAI (and eventually the HuggingFace) jars into the standard transform function

cbornet:
> It would be great to not bundle the OpenAI (and eventually the HuggingFace) jars into the standard transform function

done

pulsar-ai-functions/pom.xml
cbornet: done

pulsar-ai-functions/src/main/resources/META-INF/services/pulsar-io.yaml
cbornet: changed to `ai-tools` atm

eolivelli (approved):
Lgtm

PR #86, opened 2023-06-26
Factorize the way the result field is set

cbornet:
With this the embeddings an query can set the result as property, key field, key, value, etc...


eolivelli (approved):
I like this kind of patches that throw away lots of LOCs :-) 

+1

cbornet:
We should probably cut a release quickly with this change as it affects the API.

PR #89, opened 2023-06-28
[compute-ai-embeddings] Some enhancements to HF configuration

eolivelli:
This patch introduces some small enhancements to make the life of HF users easier:
1) Automatically detect HF service (if you configured HF and not OpenAI then it is clear that you want HF)
2) Automatically compose HF URLs when working on "local" mode
3) Allow Only some djl:// urls but default, limiting the models to sentence transformers

With this patch this is how you configure compute-ai-embeddings with HF.

API MODE:

```

{
  "datasource": {
     ...  
},
  "huggingface": {
    "provider": "api",
    "access-key": "xxxxxxxx"
    },
  "steps": [
    {
      "type": "compute-ai-embeddings",
      "model": "sentence-transformers/all-MiniLM-L6-v2",
      "embeddings-field": "value.embeddings"
      "text": "{{ value.name }} {{ value.description }}"
    },
    {
      "type": "query",
      "query": "SELECT * FROM vsearch.products ORDER BY item_vector ANN OF ? LIMIT 1;",
      "output-field": "value.results",
      "fields": ["value.embeddings"]
    }
  ]
}
```

LOCAL MODE:

```

{
  "datasource": {
     ...  
},
  "huggingface": {
    "provider": "local"
    },
  "steps": [
    {
      "type": "compute-ai-embeddings",
      "model": "sentence-transformers/all-MiniLM-L6-v2",
      "embeddings-field": "value.embeddings",
      "text": "{{ value.name }} {{ value.description }}"
    },
    {
      "type": "query",
      "query": "SELECT * FROM vsearch.products ORDER BY item_vector ANN OF ? LIMIT 1;",
      "output-field": "value.results",
      "fields": ["value.embeddings"]
    }
  ]
}
```

### DIFF
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/AbstractHuggingFaceEmbeddingService.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/AbstractHuggingFaceEmbeddingService.java
@@ -43,7 +43,7 @@ public abstract class AbstractHuggingFaceEmbeddingService<IN, OUT>
    */
   public static final String URL_PREFIXES_SYSTEM_PROP = "ALLOWED_HF_URLS";
 
-  private static final String DLJ_BASE_URL = "djl://ai.djl.huggingface.pytorch/sentence-transformers/";
+  public static final String DLJ_BASE_URL = "djl://ai.djl.huggingface.pytorch";
 
   public static final Set<String> allowedUrlPrefixes = getHuggingFaceAllowedUrlPrefixes();
 
@@ -97,11 +97,7 @@ public abstract class AbstractHuggingFaceEmbeddingService<IN, OUT>
   public AbstractHuggingFaceEmbeddingService(HuggingFaceConfig conf)
       throws IOException, ModelNotFoundException, MalformedModelException, IllegalAccessException {
     Objects.requireNonNull(conf);
-
-    if (conf.modelUrl == null) {
-      conf.modelUrl = DLJ_BASE_URL + conf.modelName;
-      log.info("Automatically computed model URL {}", conf.modelUrl);
-    }
+    Objects.requireNonNull(conf.modelName);
 
     checkIfUrlIsAllowed(conf.modelUrl);
 
--- a/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/AbstractHuggingFaceEmbeddingService.java
+++ b/pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/AbstractHuggingFaceEmbeddingService.java
@@ -43,6 +43,8 @@ public abstract class AbstractHuggingFaceEmbeddingService<IN, OUT>
    */
   public static final String URL_PREFIXES_SYSTEM_PROP = "ALLOWED_HF_URLS";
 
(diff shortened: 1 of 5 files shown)


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/AbstractHuggingFaceEmbeddingService.java
```diff
@@ -43,6 +43,8 @@ public abstract class AbstractHuggingFaceEmbeddingService<IN, OUT>
    */
   public static final String URL_PREFIXES_SYSTEM_PROP = "ALLOWED_HF_URLS";
 
+  private static final String DLJ_BASE_URL = "djl://ai.djl.huggingface.pytorch/sentence-transformers/";
```
dlg99, line 46: "djl://" should be enough. or "djl://ai.djl.huggingface.pytorch/" if you want to reuse it to build modelUrl.

"sentence-transformers" is not a tag/filter but a part of the model name in this case. E.g. something like "flax-sentence-embeddings/stackoverflow_mpnet-base" can be valid.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/AbstractHuggingFaceEmbeddingService.java
```diff
@@ -95,7 +97,11 @@ public static class HuggingFaceConfig {
   public AbstractHuggingFaceEmbeddingService(HuggingFaceConfig conf)
       throws IOException, ModelNotFoundException, MalformedModelException, IllegalAccessException {
     Objects.requireNonNull(conf);
-    Objects.requireNonNull(conf.modelUrl);
+
+    if (conf.modelUrl == null) {
+      conf.modelUrl = DLJ_BASE_URL + conf.modelName;
```
dlg99, line 102: FYI model name and the archive name aren't necessary the same but this should work in basic cases. 
Some archives can include multiple models etc. 
http://djl.ai/docs/load_model.html#customize-modelname

In this case it makes sense to set conf.modelName to null to avoid using it the way optModelName is used.

dlg99: Overall this logic should not be here. 
It belongs in TransformFunction where the service config is built from transform's config:
```java
          case LOCAL:
...
            AbstractHuggingFaceEmbeddingService.HuggingFaceConfig.HuggingFaceConfigBuilder builder =
                AbstractHuggingFaceEmbeddingService.HuggingFaceConfig.builder()
...
```

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/AbstractHuggingFaceEmbeddingService.java
eolivelli: updated

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/embeddings/AbstractHuggingFaceEmbeddingService.java
eolivelli: I have moved this code to TransformFunction.

In my testing I need to set both modelUrl and modelName, so I am not setting modeName to null

PR #90, opened 2023-06-28
Add support for JSON in compute, chat, embeddings and query

cbornet:


pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeStep.java
```diff
@@ -79,9 +81,13 @@ public void process(TransformContext transformContext) {
   }
 
   public void computeValueFields(List<ComputeField> fields, TransformContext context) {
-    if (context.getValueSchema().getSchemaInfo().getType() == AVRO) {
+    SchemaType schemaType = context.getValueSchema().getSchemaInfo().getType();
+    if (schemaType == AVRO) {
```
eolivelli, line 85: nit:
what about pushing this into a method that accepts the schemaType ? This way we will have less point in the code of the steps that deal with AVRO/JSON explicitly, and it will be easier to maintain the code (and add new steps)
`context.addOrReplaceFields(schemaType, evaluatedFields, valueSchemaCache);`

eolivelli (approved):
LGTM

I left a suggestion for a follow up work

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/ComputeStep.java
cbornet: Good catch! done.

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformContext.java
```diff
@@ -182,10 +181,21 @@ public static byte[] serializeGenericRecord(GenericRecord record) throws IOExcep
...
-  public void addOrReplaceAvroValueFields(
+  public void addOrReplaceValueFields(
+      SchemaType schemaType,
+      Map<org.apache.avro.Schema.Field, Object> newFields,
+      Map<org.apache.avro.Schema, org.apache.avro.Schema> schemaCache) {
+    if (schemaType == SchemaType.AVRO) {
+      addOrReplaceAvroValueFields(newFields, schemaCache);
+    } else if (schemaType == SchemaType.JSON) {
+      addOrReplaceJsonValueFields(newFields, schemaCache);
+    }
```
eolivelli, line 192: fail ?

eolivelli (commented):
@cbornet I have left one last comment

pulsar-transformations/src/main/java/com/datastax/oss/pulsar/functions/transforms/TransformContext.java
cbornet: I don't know. We should either fail or ignore.
I'm not sure we're consistent everywhere for this.

PR #96, opened 2023-07-05
Remove dependency on Pulsar for TransformContext and all step classes

cbornet:
This way the transformations are agnostics of the messaging system and could be extracted to a distinct module later.

### DIFF
--- a/pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/DropFieldStepTest.java
+++ b/pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/DropFieldStepTest.java
@@ -259,54 +259,6 @@ public class DropFieldStepTest {
   }
 
   @Test
-  void testAvroNotModified() throws Exception {
-    RecordSchemaBuilder recordSchemaBuilder = SchemaBuilder.record("record");
-    recordSchemaBuilder.field("firstName").type(SchemaType.STRING);
-    recordSchemaBuilder.field("lastName").type(SchemaType.STRING);
-    recordSchemaBuilder.field("age").type(SchemaType.INT32);
-
-    SchemaInfo schemaInfo = recordSchemaBuilder.build(SchemaType.AVRO);
-    GenericSchema<GenericRecord> genericSchema = Schema.generic(schemaInfo);
-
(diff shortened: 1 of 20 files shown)


pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/FlattenStepTest.java
```diff
@@ -264,21 +263,21 @@ void testNestedKeyValueNonAvroSchema() throws Exception {
...
-            "myKey");
-
-    // then
-    Utils.process(nestedKVRecord, FlattenStep.builder().part("value").build());
-  }
+//    @Test(
+//      expectedExceptions = IllegalStateException.class,
+//      expectedExceptionsMessageRegExp = "Flatten requires non-null schemas!"
+//    )
+//    void testNestedSchemalessValue() throws Exception {
```
cbornet, line 270: Can a received message schema be null ?

eolivelli: If you are using AUTO_CONSUME then it should not possibile, but I am not sure.

eolivelli (approved):
Lgtm

But I won't remove existing tests

pulsar-transformations/src/test/java/com/datastax/oss/pulsar/functions/transforms/DropFieldStepTest.java
```diff
@@ -258,54 +258,6 @@ void testKeyValueBytesJson() throws Exception {
         "{\"valueField3\":\"value3\"}");
   }
 
-  @Test
-  void testAvroNotModified() throws Exception {
```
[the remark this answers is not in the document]
cbornet: Yes. Because now  when AVRO is used, we always convert to NATIVE AVRO, even if the key/value was not modified.
Since we will serialize to bytes later, it doesn't affect performance.
But I can check if the previous behaviour can be kept if needed.

nicoloboschi (approved):

cbornet:
> But I won't remove existing tests

I merged. We can see about adding back the tests and the detection of modification later.

PR #111, opened 2023-08-02
Compute embeddigs: some small fixes

eolivelli:
Summary:
- remove the list of possible models for compute-ai-embeddings (because we listed the models for openai, now we support multiple other models and also the list of models supported by openai may change in the future)
- fix a problem with JsonConverter and ByteBuffer values

### DIFF
--- a/pulsar-ai-tools/src/main/resources/config-schema.yaml
+++ b/pulsar-ai-tools/src/main/resources/config-schema.yaml
@@ -274,10 +274,7 @@ components:
             model:
               description: ID of the model to use. See the [model endpoint compatibility](https://platform.openai.com/docs/models/model-endpoint-compatibility) table for details on which models work with the Chat API.
               example: "gpt-3.5-turbo"
-              oneOf:
-                - type: string
-                - type: string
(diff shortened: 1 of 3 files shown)


pulsar-ai-tools/src/main/resources/config-schema.yaml
```diff
@@ -274,10 +274,7 @@ components:
             model:
               description: ID of the model to use. See the [model endpoint compatibility](https://platform.openai.com/docs/models/model-endpoint-compatibility) table for details on which models work with the Chat API.
               example: "gpt-3.5-turbo"
-              oneOf:
-                - type: string
-                - type: string
-                  enum: [ "gpt-4","gpt-4-0613","gpt-4-32k","gpt-4-32k-0613","gpt-3.5-turbo","gpt-3.5-turbo-16k","gpt-3.5-turbo-0613","gpt-3.5-turbo-16k-0613" ]
+              type: string
```
cbornet, line 277: Actually OpenAI anticipated that the models could evolve. That’s why there are 2 string types, one with an enumeration and one without.
The problem is that they used oneof instead of anyof so you can’t validate both string types.

cbornet: It provide hints on the possible values to the user of the spec. We could add the Azure models to that list.

cbornet: There’s a link in the description field but I don’t think it’s used in validation exceptions.

PR #119, opened 2024-05-03
deps: Update azure-ai-openai version  to fix CVE-2023-34062

nikhilerigila09:
This PR fixes CVE-2023-34062 ( http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2023-34062)

CVE-2023-34062 is cause by
- io.projectreactor.netty:reactor-netty-http@1.0.31
- io.projectreactor.netty:reactor-netty-core@1.0.31

By upgrading azure-ai-openai version to 1.0.0-beta.6 
It upgrades the versions to 
- io.projectreactor.netty:reactor-netty-http@1.0.39
- io.projectreactor.netty:reactor-netty-core@1.0.39

### DIFF
--- a/streaming-ai/src/main/java/com/datastax/oss/streaming/ai/completions/OpenAICompletionService.java
+++ b/streaming-ai/src/main/java/com/datastax/oss/streaming/ai/completions/OpenAICompletionService.java
@@ -20,6 +20,8 @@ import static com.datastax.oss.streaming.ai.util.TransformFunctionUtil.getIntege
 
 import com.azure.ai.openai.OpenAIClient;
 import com.azure.ai.openai.models.*;
+
+import javax.management.relation.RoleNotFoundException;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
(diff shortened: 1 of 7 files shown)


streaming-ai/src/main/java/com/datastax/oss/streaming/ai/completions/OpenAICompletionService.java
```diff
@@ -66,8 +58,25 @@ public ChatCompletions getChatCompletions(
...
+      case "assistant":
+        return new ChatRequestAssistantMessage(message.getContent());
+      case "user":
+        return new ChatRequestUserMessage(message.getContent());
+      case "function":
+        return new ChatRequestFunctionMessage("", message.getContent());
+      case "tool":
+        return new ChatRequestToolMessage(message.getContent(), "");
+      default:
+        return new ChatRequestMessage();
```
nicoloboschi, line 74: should we throw error here ? it seems we're losing the content

nicoloboschi (changes requested):

streaming-ai/src/main/java/com/datastax/oss/streaming/ai/completions/OpenAICompletionService.java
nikhilerigila09: Yes, an error would be better, a message without any role or invalid role should throw an error, as there is no class which could accept the such a message.

nicoloboschi (approved):

